PR #69, opened 2014-03-04
Fix highlight error

jsvisa:
add highliht of `do:`
remove highlight of `function`

preview is 

preview

now is 

now


mydearxym:
end is still not highlight 
[Gyazo](https://gyazo.com/75c54958ba08e8490094604d71e0aa4c)

PR #70, opened 2014-03-05
Fix: no highlight of `end`

jsvisa:
Ok, I send a new issue here, just the same of [issue 69](https://github.com/elixir-lang/vim-elixir/pull/69)


mydearxym:
end is still not highlight 
[Gyazo](https://gyazo.com/75c54958ba08e8490094604d71e0aa4c)

PR #102, opened 2015-01-05
Added installation instructions for Vundle

V01D-eXe:


kassio:
I'd not like to add this, otherwise we'll need to add for `neobundle`, `pathogen`, etc. Besides this is already a pattern and the users of this tools know how to use them with any plugin.

Issue #100, opened 2014-12-28
Escaping ' in char lists not working. 


bgianfo:
For a repro, see the following example from: lib\elixir\lib\uri.ex 

      def char_reserved?(c) do
        c in ':/?#[]@!$&\'()*+,;='
      end

Seems that \' isn't being escaped properly. 

![image](https://cloud.githubusercontent.com/assets/1212/5562808/913960a6-8df1-11e4-98cc-81cabfd87777.png)

whatyouhide:
Hi @bgianfo, why did you close #101? Seemed like a fine solution to me :smiley:.

bgianfo:
Didn't work, it failed the test suite for a different feature...

Closed it as it I couldn't run the test suite on my machine at the time (
was stuck on a windows box ) and didn't have any more time to put into the
investigation.

PR #112, opened 2015-04-28
Fix highlighting of char lists with escaped quotes

flupke:


LnL7:
This seems to break [one of the tests](https://github.com/elixir-lang/vim-elixir/blob/master/spec/syntax/list_spec.rb#L4).
Could you write a test or give an example of the issue.

flupke:
Yup, I just found out about #100, another pull request was already rejected for this issue. Investigating why the tests are broken now...

LnL7:
I have a feeling the list test only passed by accident, we probably need some more tests for strings.

flupke:
Yes it looks like a typo in the test:
```
'"\'
var = 1
```
The string is not closed. I'm not sure what the test was trying to demonstrate though: `List syntax should properly handle "\\\\" inside [???]`

flupke:
Oh, I see, the backspace is not properly escaped in the test.

LnL7:
I think the test is to check if in `'\\'` the backslash is escaped instead of the quote

flupke:
Yes, so I think my fix is correct.

LnL7:
Indeed, you have to escape for ruby aswell.

PR #115, opened 2015-05-11
Add highlight for function guards.

kassio:
Highlight [the functions](http://elixir-lang.org/getting-started/case-cond-and-if.html#expressions-in-guard-clauses) that can be used in guards.

Closes #107.


kassio:
vim 2015-05-11 18-24-55

PR #116, opened 2015-07-17
Add spawn and monitor keyword synatx highlight

JackDrogon:
Add spawn and monitor keyword synatx highlight。 These functions in erlang are bifs。It's not used to using these functions  as normal function without hightlighting。


carlosgaldino:
Since these are all functions there's no point in adding them as keywords.

Thank you.

PR #141, opened 2016-02-05
Look up docs with the "K" command

padde:
This sets `iskeyword` to include `.`, so that expressions of the form `Module.function` will be treated as a whole when looking up documentation with the `K` command.

The matched expression is then passed as command line argument to the `keywordprg` which is defined as:

```
iex -e 'Code.eval_string("import IEx.Helpers; h " <> hd(System.argv)); System.halt(0)' -- {MATCHED_EXPRESSION}
```

So if we press `K` in normal mode, we will see the documentation for the module or function under the cursor, as it is shown in IEx.

### DIFF
--- a/ftplugin/elixir.vim
+++ b/ftplugin/elixir.vim
@@ -18,3 +18,6 @@ endif
 
 setlocal comments=:#
 setlocal commentstring=#\ %s
+
+setlocal iskeyword+=.
+setlocal keywordprg=iex\ -e\ 'Code.eval_string(\"import\ IEx.Helpers;\ h\ \"\ <>\ hd(System.argv));\ System.halt(0)'\ --


ftplugin/elixir.vim
```diff
@@ -18,3 +18,6 @@ endif
 
 setlocal comments=:#
 setlocal commentstring=#\ %s
+
+setlocal iskeyword+=.
```
kassio: I don't know if I agree with this, since this will change the autocomplete behaviour too.

padde: @kassio you're right, this is not good. It will change the behavior of `*` as well. I'll check if this can be solved differently.

As a last resort it could be solved by remapping `K` to a custom command. I think it would be nice to make these commands usable in elixir mode, because they are useless with Vim's default configuration anyway. Same thing applies to the `gf` command. The more I think about it, the more I embrace the idea of remapping it such that it will open a `.ex` file instead of the directory with the same name. What do you think about remapping standard commands in general?

aerosol: FWIW https://github.com/aerosol/dotfiles/blob/develop/silos/nvim/.config/nvim/init.vim#L397-L450

kassio:
I'll close it for now.

padde:
👍

PR #152, opened 2016-04-24
TestFile command to run one test file inside vim

RobertDober:
I was just hacking up this quick hack to scratch a very personal itch. As I could not figure out how to do this with the current code in  `compiler/exunit.vim` I thought sharing this approach might be a good idea.

Is this a sensible idea?

Thx in advance for your feedback.

### DIFF
--- a/autoload/elixir/test.vim
+++ b/autoload/elixir/test.vim
@@ -1,0 +1,35 @@
+let s:awk_filter =  ['{gsub("^ *","")}']
+let s:awk_filter += ['/^Assertion/{a=$0}']
+let s:awk_filter += ['/^code:/{gsub("code: *","");c=$0}']
+let s:awk_filter += ['/^stacktrace:/{getline; gsub("^ *",""); print $0 ":" a " " c}']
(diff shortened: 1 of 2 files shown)


kassio:
I don't know if this is a good idea. I rather keep this repo for indenting and syntax stuff. I'll leave this open and see if we get more opinions here, wdyt?

RobertDober:
Sure, I believe in small (Unix) like specialized  tools that do their job well, so I completely understand.
Maybe I can come up with a usable vim-elixir-test plugin ;).

aerosol:
@RobertDober https://github.com/janko-m/vim-test works great

RobertDober:
@aerosol interesting, however in the meantime I have taken tmux into the receipt and I am sending tmux commands from my vim to the test pane and switch to it too, that suits my workflow; However I will ckeck your plugin with Neovim mode as soon as possible.

kassio:
Well, I think I can close it now. Besides `vim-test`, with neovim you could try [neoterm](http://github.com/kassio/neoterm) too.

PR #158, opened 2016-06-07
Use elixirModuleDeclaration for highlights

vsushkov:
I'm trying to update the Tomorrow colorscheme, so it better matches syntax highlighting in official elixir documentation. One of the challenges I've met is I want module name in `defmodule` statement to
have aqua color and when it's not in `defmodule`, it must be orange.
screen shot 2016-06-07 at 19 28 52
As I understand, it's impossible to do with current syntax. So I suggest this pull request. If you think it's a good idea, then I'll do the same for [all remaining matchers for declarations](https://github.com/elixir-lang/vim-elixir/pull/158).
Before this commit module name is highlighted as `elixirAlias`, and after this commit it's recognised as `elixirModuleDeclaration`.
It's very helpful to make your vim colorscheme more customisable.

### DIFF
--- a/spec/syntax/demodule_spec.rb
+++ b/spec/syntax/demodule_spec.rb
@@ -1,0 +1,33 @@
+require 'spec_helper'
+
+describe "Defmodule syntax" do
+  it "defines `defmodule` keyword as elixirModuleDefine" do
+    <<-EOF
+      defmodule HelloPhoenix.HelloController do
+    EOF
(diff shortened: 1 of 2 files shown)


vsushkov:
@kassio does this merge means it's a good idea to do the same for all remaining matches and declarations?

kassio:
Yeah, go for it. :)

PR #171, opened 2016-09-02
After a line that starts with a pipe operator we always want to stay …

offthecuff:
…at the same indentation.


emilsoman:
@offthecuff why did you close this ?

offthecuff:
I jumped the gun on this. I opened it with test cases not passing, etc. I ended up submitting this issue instead https://github.com/elixir-lang/vim-elixir/issues/173.

PR #185, opened 2016-10-14
Test to show @doc with sigil_S don't show as elixirDocString.

lmarlow:
A lot of Elixir's documentation is written with `sigil_S` for `@moduledoc`s and function `@doc`s. Currently they are all matched as `elixirSigil`, instead of `elixirDocString`.

I haven't been able to fix it, so I thought I'd submit a PR with the failing test case.

### DIFF
--- a/spec/syntax/heredoc_spec.rb
+++ b/spec/syntax/heredoc_spec.rb
@@ -12,7 +12,7 @@ describe 'Heredoc syntax' do
       EOF
     end
 
-    pending 'doc with sigil_S triple double-quoted multiline content' do
+    it 'doc with sigil_S triple double-quoted multiline content' do
       expect(<<~EOF).to include_elixir_syntax('elixirDocString', 'foo')
         @doc ~S"""
         foo
@@ -20,19 +20,11 @@ describe 'Heredoc syntax' do
       EOF
     end
 
-    pending 'doc with sigil_S triple single-quoted multiline content' do
+    it 'doc with sigil_S triple single-quoted multiline content' do
       expect(<<~EOF).to include_elixir_syntax('elixirDocString', 'foo')
         @doc ~S'''
         foo
         '''
-      EOF
-    end
-
-    it 'doc escapes quotes unless only preceded by whitespace' do
-      expect(<<~EOF).to include_elixir_syntax('elixirDocString', %q(^\s*\zs"""))
-        @doc """
-        foo """
-        """
       EOF
     end
 
--- a/spec/syntax/heredoc_spec.rb
+++ b/spec/syntax/heredoc_spec.rb
@@ -4,7 +4,7 @@ require 'spec_helper'
 
 describe 'Heredoc syntax' do
   describe 'binary' do
-    it 'with multiline content' do
+    it 'doc with multiline content' do
       expect(<<~EOF).to include_elixir_syntax('elixirDocString', 'foo')
         @doc """
         foo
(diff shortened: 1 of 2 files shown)


spec/syntax/heredoc_spec.rb
```diff
@@ -4,28 +4,42 @@
...
       expect(<<~EOF).to include_elixir_syntax('elixirDocString', 'foo')
         @doc """
         foo
         """
       EOF
     end
 
-    it 'escapes quotes unless only preceded by whitespace' do
+    it 'doc with sigil_S' do
+      expect('~S(foo)').to include_elixir_syntax('elixirDocString', 'foo')
```
kassio: this must not include `elixirDocString` since it doesn't begins with `@doc`, right?

lmarlow: True. With `@doc` it would be a valid `elixirDocString`, but it's also not a heredoc, so I'm not sure it belongs in this test file. I haven't found any examples of docs being written that way. I have found some with `~S'''`, though. I'll modify the test.

kassio:
The tests you added are breaking in CI, can you take a look?

lmarlow:
Would you prefer pending tests? I haven't been able to fix the issue myself, so I was hoping someone with better vim syntax skills could fix them.

kassio:
Yes, please, add them as `pendent`. And, also, open an issue. This way will be easier to keep track of what I have to fix. 😄

lmarlow:
I'm not sure why these [tests](https://travis-ci.org/elixir-lang/vim-elixir/builds/169312914) are failing. They seem to be unrelated and were evening failing when my newly added specs were pending. They all pass locally.

kassio:
So, I think the desired result should be something like this:
python
Note that elixir does not uses `'''` for documentation, only `"""`. Also, I rather keep the `@*doc` as `elixirVariable`.

However, with your code the result is this:
python

IMHO, I think we could add the `~S` as an optional part of the current `elixirDocString` regex, like this: 
```viml
syn match elixirDocString +\(@\w*doc\s*\)\@<=\%(\(\~[Ss]\)\?"""\_.\{-}\_^\s*"""\|".\{-\}"\)+ contains=elixirTodo,elixirInterpolation,@Spell fold
```

Also, to make this match works we must move this line to bellow the `elixirSigil` matches, after line 113, otherwise the `elixirSigil` region will override it.

Thank you for the PR, I really appreciate that.  ❤️💛💚💜💙
Sorry for the delay in the help.

lmarlow:
This works now to keep the `elixirVariable` for `@doc` which is better. It's still not perfect, as the delimiters around the doc string are included in the doc string.

Thanks for your help. I think I went this way initially, but didn't know I had to escape `~` in the regex.

kassio:
My unique problem with that now is that it's recognising the `'''` as documentation, while elixir does not do that:
```console
/tmp
$ cat a.ex
defmodule A do
  @doc '''
  documentation
  '''
  @doc """
  documentation
  """
  @doc ~S"""
  documentation
  """
  @doc ~S'''
  documentation
  '''
end
/tmp
$ iex a.ex
Erlang/OTP 19 [erts-8.1] [source] [64-bit] [smp:4:4] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

** (ArgumentError) expected the doc attribute to contain a binary, a boolean, or nil, got: 'documentation\n'
    (elixir) lib/module.ex:1078: Module.preprocess_attribute/2
    (elixir) lib/module.ex:860: Module.put_attribute/4
    a.ex:2: (module)
    (stdlib) erl_eval.erl:670: :erl_eval.do_apply/6
```

Can you remove the `\|'''` from the regex, and its tests?

lmarlow:
Elixir uses the `~S'''` in one place https://github.com/elixir-lang/elixir/blob/8fa5b6953efab37b5c119daaf82cf270ca1772bb/lib/elixir/lib/module.ex#L2. It doesn't seem common. Would you like me to try to match them all or just `"""` and `~S"""` for the common cases for now?

kassio:
humm.. that interesting.. it accepts `~S'''`, but does not accept it without the sigil. 😢 

You were right since the begining, so it'll be better to use regions than match. What do you think to use this solution:
```viml
syn region elixirDocString matchgroup=elixirDocString start=+\(@\w*doc\s\+\)\@<=\z("""\)+ end=+\z1+ contains=elixirTodo,elixirInterpolation,@Spell fold
syn region elixirDocString matchgroup=elixirDocString start=+\(@\w*doc\s\+\)\@<=\~[Ss]\z('''\|"""\)+ end=+\z1+ contains=elixirTodo,elixirInterpolation,@Spell fold
```

This produces:
python

kassio:
Also, if you don't mind, would be good to have a test validating that `@doc '''` must not highlight as documentation, wdyt?

lmarlow:
Test added and delimiters are more visible now.

PR #193, opened 2016-10-21
adds support for vimrunners other than gvim

jbodah:
Adds support for running tests with other vim derivatives. Defaults to gvim, then looks for mvim, and finally tries vanilla vim


jbodah:
Turns out this just works without this change if I install mvim. Closing

PR #203, opened 2016-10-26
Use markdown syntax for `elixirDocString`

lmarlow:
Enables seeing code samples and spellchecking docs more easily when `g:elixir_use_markdown_for_docs` is set.

image

### DIFF
--- a/spec/syntax/doc_spec.rb
+++ b/spec/syntax/doc_spec.rb
@@ -2,8 +2,18 @@
 
 require 'spec_helper'
 
-describe 'Heredoc syntax' do
-  describe 'binary' do
+describe 'documentation syntax' do
+  describe 'string' do
+    it 'doc in double quotes' do
+      expect('@doc "foo"').to include_elixir_syntax('elixirDocString', 'foo')
+    end
+
+    it 'doc in sigil_S' do
+      expect('@doc ~S(foo)').to include_elixir_syntax('elixirDocString', 'foo')
+    end
+  end
+
+  describe 'heredoc' do
     it 'doc with multiline content' do
       ex = <<~EOF
         @callbackdoc """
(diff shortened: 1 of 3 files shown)


kassio:
Is `markdown.vim` part of vim default runtime? If it isn't I think it would be better check if it exists before highlighting with it and, also, fallback to the old `ElixirDocString` if it doesn't exists in the current vim instance. WDYT?

jbodah:
What about making this configurable via a variable in .vimrc? Sounds like that could solve the question of `markdown.vim` not being installed and also allow people to choose whether they want markdown or string syntax

kassio:
The configurable solution sounds good, wdyt @lmarlow ?

lmarlow:
Markdown has been included with vim since 7.3 in 2010.

I'm fine with making it configurable. I'll look around for examples as my _vimscript-fu_ is still weak.

kassio:
@lmarlow cool, thanks! If you need help ping me on slack. 😄

lmarlow:
@kassio, can you take a look at this? I think it's ready now.

kassio:
It looks nice, but there is a broken test on CI, can you check it?

lmarlow:
The failure seemed order dependent. I pushed a change that puts setting the `g:elixir_use_markdown_for_docs` into its own `describe` block.

lmarlow:
👍 Thanks!

pshoukry:
if `let g:elixir_use_markdown_for_docs=0` or the setting isn't set in .vimrc it leaks to the following method.
no properly parsed

lmarlow:
@pshoukry, does it look right if you add a newline after the `""` line? I've always seen `iex>` sections end with a blank line after them.

pshoukry:
@lmarlow Yup that fixed it.

PR #205, opened 2016-10-27
Add support for vim-surround

### LINKED ISSUE #197: Add support for vim-surround for EEx
It would be something like:

```vim
if !exists('b:surround_45')
  let b:surround_45 = "<% \r %>"
endif
if !exists('b:surround_61')
  let b:surround_61 = "<%= \r %>"
endif
if !exists('b:surround_35')
  let b:surround_35 = "<%# \r %>"
endif
if !exists('b:surround_5')
  let b:surround_5 = "<% \r %>\n<% end %>"
endif
```

Extracted from https://github.com/avdgaag/vim-phoenix

hauleth:
Close #197

### DIFF
--- a/ftplugin/eelixir.vim
+++ b/ftplugin/eelixir.vim
@@ -79,15 +79,19 @@ if exists("loaded_matchit")
 endif
 
 if !exists('b:surround_45')
+  " When using surround `-` (ASCII 45) would provide `<% selection %>`
   let b:surround_45 = "<% \r %>"
 endif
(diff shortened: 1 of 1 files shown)


kassio:
Can you add some comments in the code explaining what is that? I mean, what this numbers mean and a reference for the surround plugin, please?

Thanks,

hauleth:
Added.

kassio:
Thanks!

Issue #206, opened 2016-10-28
Wrong colour for keyword "length" in keyword list


brianvanburken:
When adding a keyword "length" it gets coloured in red. 
Expected colour is same as keyword above.

![screen shot 2016-10-28 at 10 42 32](https://cloud.githubusercontent.com/assets/1044316/19800342/9505efe0-9cfb-11e6-8e07-1544715806b0.png)

jbodah:
Hi, are you using the latest master? The following test is passing for me:

```rb
# frozen_string_literal: true

require 'spec_helper'

describe 'Keyword syntax' do
  it 'a kernel function used as an atom key in a keyword list' do
    expect(<<~EOF).to include_elixir_syntax('elixirAtom', 'length')
    plug Plug.Parsers,
      parsers: [:urlencoded, :multipart, :json],
      pass: ["*/*"],
      json_decoder: Poison,
      length: 400_000_000
    EOF
  end
end
```

brianvanburken:
I've checked and yes I'm using the latest master. According to the local repo I'm on commit d3d964a8b1953e12cdadb77b3f255923f4873637 for master. I've tried to debug if other plugins are causing this but no result. When only enabling the `vim-elixir` the problem still exists. So far I'm able to reproduce this on my machine.

This is my .vimrc file: https://github.com/brianvanburken/dotfiles/blob/master/.vimrc
The file I'm testing the problem with: https://github.com/brianvanburken/elixir-playground/blob/master/phoenix_video_stream/lib/phoenix_video_stream/endpoint.ex

My Vim info:

```
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Oct 22 2016 08:48:49)
MacOS X (unix) version
Included patches: 1-46
Compiled by Homebrew
Huge version without GUI.  Features included (+) or not (-):
+acl             -ebcdic          +listcmds        +persistent_undo +textobjects
+arabic          +emacs_tags      +localmap        +postscript      +timers
+autocmd         +eval            -lua             +printer         +title
-balloon_eval    +ex_extra        +menu            +profile         -toolbar
-browse          +extra_search    +mksession       +python          +user_commands
++builtin_terms  +farsi           +modify_fname    -python3         +vertsplit
+byte_offset     +file_in_path    +mouse           +quickfix        +virtualedit
+channel         +find_in_path    -mouseshape      +reltime         +visual
+cindent         +float           +mouse_dec       +rightleft       +visualextra
[...]

PR #207, opened 2016-10-28
fixes Wrong colour for keyword "length" in keyword list #206

jbodah:
fixes Wrong colour for keyword "length" in keyword list #206

Note that this also removes highlighting in some places that I found it
to be useful (e.g. using `is_pid` in a function body) but that can be
discussed in a different PR

### DIFF
--- a/syntax/elixir.vim
+++ b/syntax/elixir.vim
@@ -22,13 +22,9 @@ syn match elixirKeyword '\(\.\)\@<!\<\(for\|case\|when\|with\|cond\|if\|unless\|
 syn match elixirKeyword '\(\.\)\@<!\<\(exit\|raise\|throw\|after\|rescue\|catch\|else\)\>'
 syn match elixirKeyword '\(\.\)\@<!\<\(quote\|unquote\|super\|spawn\|spawn_link\|spawn_monitor\)\>'
 
-" Functions used on guards
-syn keyword elixirKeyword contained is_atom is_binary is_bitstring is_boolean
-syn keyword elixirKeyword contained is_float is_function is_integer is_list
-syn keyword elixirKeyword contained is_map is_nil is_number is_pid is_port
-syn keyword elixirKeyword contained is_record is_reference is_tuple is_exception
-syn keyword elixirKeyword contained abs bit_size byte_size div elem hd length
-syn keyword elixirKeyword contained map_size node rem round tl trunc tuple_size
+" Kernel functions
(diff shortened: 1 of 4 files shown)


syntax/elixir.vim
```diff
@@ -22,13 +22,9 @@ syn match elixirKeyword '\(\.\)\@<!\<\(for\|case\|when\|with\|cond\|if\|unless\|
...
 
-" Functions used on guards
-syn keyword elixirKeyword contained is_atom is_binary is_bitstring is_boolean
-syn keyword elixirKeyword contained is_float is_function is_integer is_list
-syn keyword elixirKeyword contained is_map is_nil is_number is_pid is_port
-syn keyword elixirKeyword contained is_record is_reference is_tuple is_exception
-syn keyword elixirKeyword contained abs bit_size byte_size div elem hd length
-syn keyword elixirKeyword contained map_size node rem round tl trunc tuple_size
+" Kernel functions
+syn match elixirKernelFunction contained containedin=elixirGuard '\<is_atom\|is_binary\|is_bitstring\|is_boolean\|is_float\|is_function\|is_integer\|is_list\|is_map\|is_nil\|is_number\|is_pid\|is_port\>\([ (]\)\@='
```
kassio: I think we should use a group on these regex: `\<\(is_atom\|...`, right?

jbodah:
Here's a simplified version of the problem:

test_ex______repos_vim-elixir__-_vimrunner0_4636016625788616

And here's this branch:

test_ex______repos_vim-elixir__-_vimrunner0_42362569624178137

syntax/elixir.vim
jbodah: I thought so too, but it doesn't seem to make a difference from what I can tell. Maybe there's an edge case I'm missing?

kassio: In theory without the group the `\<` and `\>` are not applied to each word in the list. So, this: `\<\(A\|B\)\>` is equal to `\(\<A\>\|\<B\>\)`, whilst `\<A\|B\>` isn't.

kassio: @jbodah are you going to make this change or should I take this one over?

jbodah: fixed

jbodah:
@kassio should be all set now

kassio:
Just fix the conflicts with master and I'll hit the green button.

jbodah:
@kassio rebased

PR #208, opened 2016-10-28
fixes Indentation on piping long string #200

### LINKED ISSUE #200: Indentation on piping long string
Actual:

``` ex
  def build_command(input, output) do
    "really long command with a bunch of options -async 1 -opt int=72:min-dkjfslf=72:sdfdsfs=-1"
                                                                   |>
  end
```

Expected:

``` ex
  def build_command(input, output) do
    "really long command with a bunch of options -async 1 -opt int=72:min-dkjfslf=72:sdfdsfs=-1"
    |>
  end
```

jbodah:
* fixes #200

### DIFF
--- a/autoload/elixir/indent.vim
+++ b/autoload/elixir/indent.vim
@@ -22,6 +22,7 @@ let s:PAIR_START = '\<\%('.s:NO_COLON_BEFORE.s:BLOCK_START.'\)\>'.s:NO_COLON_AFT
 let s:PAIR_MIDDLE = '^\s*\%('.s:BLOCK_MIDDLE.'\)\>'.s:NO_COLON_AFTER.'\zs'
 let s:PAIR_END = '\<\%('.s:NO_COLON_BEFORE.s:BLOCK_END.'\)\>\zs'
 let s:LINE_COMMENT = '^\s*#'
+let s:MATCH_OPERATOR = '[^!><=]=[^~=>]'
 
 function! s:pending_parenthesis(line)
   if a:line.last.text !~ s:ARROW
@@ -162,12 +163,13 @@ endfunction
 
 function! elixir#indent#indent_pipeline_assignment(ind, line)
   if a:line.current.text =~ s:STARTS_WITH_PIPELINE
-        \ && a:line.last.text =~ '^[^=]\+=.\+$'
+        \ && a:line.last.text =~ s:MATCH_OPERATOR
     let b:old_ind.pipeline = indent(a:line.last.num)
     " if line starts with pipeline
     " and last line is an attribution
     " indents pipeline in same level as attribution
-    return match(a:line.last.text, '=\s*\zs[^ ]')
+    let assign_pos = match(a:line.last.text, '=\s*\zs[^ ]')
+    return (elixir#util#is_indentable_at(a:line.last.num, assign_pos) ? assign_pos : a:ind)
   else
     return a:ind
   end
--- a/spec/indent/pipeline_spec.rb
+++ b/spec/indent/pipeline_spec.rb
@@ -81,4 +81,42 @@ describe 'Indenting pipeline' do
     end
     EOF
   end
+
+  it 'correctly indents pipelines on a string with "=" embedded in it' do
+    expect(<<~EOF).to be_elixir_indentation
+    def build_command(input, output) do
+      "embedded=here"
+      |>
+    end
+    EOF
+  end
+
+  it 'correctly indents pipelines on a charlist with "=" embedded in it' do
+    expect(<<~EOF).to be_elixir_indentation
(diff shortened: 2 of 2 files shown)


kassio:
Instead of changing the pattern, what do you think to use the `s:is_indentable_match()` to verify if the `=` isn't a string?

jbodah:
> Instead of changing the pattern, what do you think to use the s:is_indentable_match() to verify if the = isn't a string?

That sounds good. I'm not up to speed with all of the indent code yet. I'll give that a shot and post back with how it works

kassio:
any updates on this?

jbodah:
@kassio played with your idea a bit:

```vim
function! elixir#indent#indent_pipeline_assignment(ind, line)
  if a:line.current.text =~ s:STARTS_WITH_PIPELINE
        \ && elixir#util#is_indentable_match(a:line.last, s:LAST_NON_WHITESPACE_CHAR)
        \ && a:line.last.text =~ '^[^=]\+=.\+$'
    let b:old_ind.pipeline = indent(a:line.last.num)
    " if line starts with pipeline
    " and last line is an attribution
    " indents pipeline in same level as attribution
    return match(a:line.last.text, '=\s*\zs[^ ]')
  else
    return a:ind
  end
endfunction
```

But in testing I got issues like below

```
        def test do
          [ h | t ] = "a,b,c,d"
       -              |> String.split(",")
       -              |> Enum.reverse
       +  |> String.split(",")
       +  |> Enum.reverse

       -  { :ok, h }
       +{ :ok, h }
        end
```

jbodah:
Actually, think I see what's wrong...

jbodah:
@kassio how would you feel about doing something like:

1. find the last `=` in the previous line
2. check the syntax there
2a. if match operator, then return index of operator
2b. else, look at the next `=` in previous line
3. if none found, return original indent

kassio:
I was testing it locally, and I think that maybe this can solve the problem:
```viml
function! elixir#indent#indent_pipeline_assignment(ind, line)
  if a:line.current.text =~ s:STARTS_WITH_PIPELINE
        \ && a:line.last.text =~ '^[^=]\+=.\+$'
    let b:old_ind.pipeline = indent(a:line.last.num)
    " if line starts with pipeline
    " and last line is an attribution
    " indents pipeline in same level as attribution
    let assign_pos = match(a:line.last.text, '=\s*\zs[^ ]')
    return (elixir#util#is_indentable_at(a:line.last.num, assign_pos) ? assign_pos : a:ind)
  else
    return a:ind
  end
endfunction
```

kassio:
wdyt?

jbodah:
That looks better than what I had, but it still is open to weird edge cases (e.g. multiple `=`) and map arrows as below:

```ex
  it 'correctly indents pipelines on a map with "=>" syntax' do
    expect(<<~EOF).to be_elixir_indentation
    def build_command(input, output) do
      %{:hello => :world}
      |>
    end
    EOF
  end
```

What about changing the regex there to ignore arrows too?

kassio:
may I assume that it is an assign if, and only if, the `=` sign has anything different of `<,>,=,+,-` before/after and/or it's not a string?

jbodah:
Looking [here](http://elixir-lang.org/docs/master/elixir/operators.html) all of the following have `=`:

```ex
<=
>=
==
!= 
=~ 
=== 
!==
=
=>
```

The only other cases I can think of we'd want to ignore are strings and comments (I believe both are covered now)

kassio:
@jbodah agree with you, can you update the regex with these values then?

jbodah:
@kassio let me know how this looks

autoload/elixir/indent.vim
```diff
@@ -22,6 +22,7 @@ let s:PAIR_START = '\<\%('.s:NO_COLON_BEFORE.s:BLOCK_START.'\)\>'.s:NO_COLON_AFT
 let s:PAIR_MIDDLE = '^\s*\%('.s:BLOCK_MIDDLE.'\)\>'.s:NO_COLON_AFTER.'\zs'
 let s:PAIR_END = '\<\%('.s:NO_COLON_BEFORE.s:BLOCK_END.'\)\>\zs'
 let s:LINE_COMMENT = '^\s*#'
+let s:MATCH_OPERATOR = '[^!><=]=[^~=>]'
```
kassio: Isn't this the `assign_operator`?

jbodah: Technically it's the match operator I believe, but I'm up for calling it anything - let me know which you'd prefer http://elixir-lang.org/getting-started/pattern-matching.html#the-match-operator

kassio: Oh, my bad. Totally forgot that this is elixir. 🏃

PR #214, opened 2016-10-30
add bin/spawn_vim for interactive debugging

jbodah:
This adds `bin/spawn_vim` which will spawn a new vim instance for playing around in with these configs


kassio:
❤️💛💚💜💙

PR #217, opened 2016-11-01
Move the indentation rules to autoload functions

kassio:
The main reason of this is try to simplify the rules definition.

- Simplify the `indent/elixir.vim` moving all the indentation rules to
`autoload/indent.vim`. The indent script is responsible now for call the
rules, not define them.
- Rename `GetElixirIndent` to `elixir#indent`.
- Use constants with SCREAM_CASE.
- Each rule function now receives the current indentation and a line
dictionary, which contains the current and last lines number and texts.

@jbodah can you gimme your 2 cents here?

### DIFF
--- a/autoload/elixir/indent.vim
+++ b/autoload/elixir/indent.vim
@@ -1,0 +1,208 @@
+let s:NO_COLON_BEFORE = ':\@<!'
+let s:NO_COLON_AFTER = ':\@!'
+let s:ENDING_SYMBOLS = '\]\|}\|)'
+let s:STARTING_SYMBOLS = '\[\|{\|('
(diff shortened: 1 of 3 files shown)


jbodah:
No strong opinion from me either way. Looks cleaner :+1:

sumerman:
Breaks (disables) indentation completely when is used with Vundle master.

kassio:
@sumerman can you create an issue please?

sumerman:
@kassio I was unsuccessfully trying to create a minimal example. It turns out the problem was present only in one stored session. Therefore it's not an actual issue. My apologies.

Issue #216, opened 2016-10-31
`import [defrecord: ...]` does not highlight properly


jbodah:
In the following `defrecord` is highlighted as a keyword instead of as an atom

```ex
defmodule XmlElement do
  require Record
  import Record, only: [defrecord: 2, extract: 2]
end
```

kassio:
Can you add screenshots please? And/or a failing test?

jbodah:
<img width="718" alt="test_ex______repos_vim-elixir__-_vimrunner0_04275292178254475" src="https://cloud.githubusercontent.com/assets/6093002/19890000/8aabc4f0-a00e-11e6-904e-ac1fba580ec8.png">

PR #219, opened 2016-11-01
treat keywords used as atoms as atoms

jbodah:
Addresses https://github.com/elixir-lang/vim-elixir/issues/216

### DIFF
--- a/spec/syntax/atom_spec.rb
+++ b/spec/syntax/atom_spec.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'Atom syntax' do
+  KEYWORDS = %w(
+    def
+    defp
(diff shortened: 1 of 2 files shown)


spec/syntax/atom_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    defmacrop
+    defdelegate
+    defoverridable
+    defexception
+    defcallback
+    defstruct
+  )
+
+  it '`atom:` style keyword used as an atom' do
+    KEYWORDS.each do |kw|
```
jbodah: @kassio not sure what you think of doing things like this in tests. I tend to like them for increasing test coverage, but I'm willing to split this out if you'd rather

kassio: no problem with that.

PR #221, opened 2016-11-07
fix poor indent performance on files over 1k lines

gasparch:
While editing very big file (2_500 lines, exunit test file with lots of cases), vim starts to crawl to death. I literally need to wait a few seconds on pressing Enter each time.

Profiling shows that:
```
FUNCTIONS SORTED ON SELF TIME
count  total (s)   self (s)  function
  155              6.407817  elixir#util#is_indentable_at()
  140              0.008775  <SNR>92_match_count()
   13   6.442832   0.005395  elixir#indent()
```

Turning sync off makes vim run at expected speed. In order to offer compromise between correct syntax detection and speed - minlines/maxlines are used.

### DIFF
--- a/autoload/elixir/util.vim
+++ b/autoload/elixir/util.vim
@@ -7,8 +7,9 @@ function! elixir#util#is_indentable_at(line, col)
   " Vim is making some mess on parsing the syntax of 'end', it is being
   " recognized as 'elixirString' when should be recognized as 'elixirBlock'.
   call synID(a:line, a:col, 1)
-  " This forces vim to sync the syntax.
-  syntax sync fromstart
+  " This forces vim to sync the syntax. Using fromstart is very slow on files
+  " over 1k lines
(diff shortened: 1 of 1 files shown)


kassio:
Can you add a profiling of the change? How better is that? (just curious)

gasparch:
Now delay is not noticeable :)

```
FUNCTIONS SORTED ON SELF TIME
count  total (s)   self (s)  function
   13              0.037978  elixir#util#is_indentable_at()
   12              0.000740  <SNR>95_match_count()
    1   0.043173   0.000599  elixir#indent()
    1   0.039324   0.000433  <SNR>90_is_indentable_line()
    6   0.001332   0.000260  elixir#util#count_indentable_symbol_diff()
   12   0.000332   0.000213  elixir#util#is_indentable_match()
    1              0.000081  <SNR>90_build_line()
```

one more performance update follows :)

PR #222, opened 2016-11-07
Improve indent performance

gasparch:
avoid doing extra work in indent function.

`elixir#util#is_indentable_at` guaranteed will return `true` if match is not found, so detect and do it as fast as we can. 

By calling `elixir#util#is_indentable_match` two unnecessary calls to getline() are made on each `elixir#util#count_indentable_symbol_diff` invocation, which is called at least 3 times for each line for `()[]{}` pairs. Use already fetched line content.

### DIFF
--- a/autoload/elixir/util.vim
+++ b/autoload/elixir/util.vim
@@ -2,6 +2,9 @@ let s:SKIP_SYNTAX = '\%(Comment\|String\)$'
 let s:BLOCK_SKIP = "synIDattr(synID(line('.'),col('.'),1),'name') =~? '".s:SKIP_SYNTAX."'"
 
 function! elixir#util#is_indentable_at(line, col)
+  if a:col == -1 " skip synID lookup for not found match
+    return 1
+  end
   " TODO: Remove these 2 lines
   " I don't know why, but for the test on spec/indent/lists_spec.rb:24.
   " Vim is making some mess on parsing the syntax of 'end', it is being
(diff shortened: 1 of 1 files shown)


autoload/elixir/util.vim
```diff
@@ -2,25 +2,29 @@ let s:SKIP_SYNTAX = '\%(Comment\|String\)$'
...
   " TODO: Remove these 2 lines
   " I don't know why, but for the test on spec/indent/lists_spec.rb:24.
   " Vim is making some mess on parsing the syntax of 'end', it is being
   " recognized as 'elixirString' when should be recognized as 'elixirBlock'.
   call synID(a:line, a:col, 1)
-  " This forces vim to sync the syntax.
-  syntax sync fromstart
+  " This forces vim to sync the syntax. Using fromstart is very slow on files
+  " over 1k lines
+  syntax sync minlines=20 maxlines=150
```
kassio: this was already added on master

autoload/elixir/util.vim
```diff
@@ -2,25 +2,29 @@ let s:SKIP_SYNTAX = '\%(Comment\|String\)$'
 let s:BLOCK_SKIP = "synIDattr(synID(line('.'),col('.'),1),'name') =~? '".s:SKIP_SYNTAX."'"
 
 function! elixir#util#is_indentable_at(line, col)
+  if a:col == -1 " skip synID lookup for not found match
+	  return 1
```
kassio: the indentation inside the block must be 2 spaces

autoload/elixir/util.vim
```diff
@@ -2,25 +2,29 @@ let s:SKIP_SYNTAX = '\%(Comment\|String\)$'
...
+  " over 1k lines
+  syntax sync minlines=20 maxlines=150
 
   return synIDattr(synID(a:line, a:col, 1), "name")
         \ !~ s:SKIP_SYNTAX
 endfunction
 
-function! elixir#util#is_indentable_match(line, pattern)
-  return elixir#util#is_indentable_at(a:line, match(getline(a:line), a:pattern))
+function! elixir#util#is_indentable_match(line, line_text, pattern)
```
kassio: Since, now, we have the `line` dictionary, with `:num, :text` keys, I think we could use that dictionary here, instead of add one argument, wdyt?

gasparch:
okay, now should be good. fixed spacing and it passes now single argument.

gasparch:
Do I need to do any other changes on this branch or it is ready for merge?

PR #223, opened 2016-11-07
Skip vim folds when indenting

gasparch:
using vim in foldmethod=marker mode causes problems with code

```
def foo() do # {{{
  bar()
end # }}}
```

pressing `O` on line `bar()` creates wrong indent because of {{{ mistakenly taken to be part of tuple.

```
def foo() do # {{{
  bar()
end # }}}
```

specifically mark them as `elixirFold` syntax group and skip them when detecting indent level.

### DIFF
--- a/autoload/elixir/util.vim
+++ b/autoload/elixir/util.vim
@@ -1,4 +1,4 @@
-let s:SKIP_SYNTAX = '\%(Comment\|String\)$'
+let s:SKIP_SYNTAX = '\%(Comment\|String\|elixirFold\)$'
 let s:BLOCK_SKIP = "synIDattr(synID(line('.'),col('.'),1),'name') =~? '".s:SKIP_SYNTAX."'"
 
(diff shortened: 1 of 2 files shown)


kassio:
I'm not sure we should add this code. This is too specific to vim, and to be honest I don't suggest anyone to use vim fold marks in any language but `vim-script`.
Just to understand better, do you really commit code with these markers?

gasparch:
I personally do :) may be it is not a best practice, yep.

Why I'm happy with them - it allows to fold all clauses of functions together, and not one by one. So, more logical folds, not some vim plugin decided to do folding.

gasparch:
May be I can contribute with something like http://twinside.github.io/vim-haskellFold/

kassio:
@gasparch interesting. Yeah, IMHO i think this is a little out of the scope of this plugin. But the community can appreciate a new plugin for this. 😄

PR #232, opened 2016-12-02
`elixirDocString`'s own their ends.

### LINKED ISSUE #230: Broken highlithing
in many different places of Elixir stdlib or in deps like `ecto` the highlithing is broken
This was checked in `neovim` and `vim`(with only installed `vim-elixir`) 
h

hi

lmarlow:
This prevents doc tests leaking past the end of the docstring.

Fixes #230


tsubery:
Is there a reason not to add keepend parameter to all 9 definitions of elixirDocStringContained region?

lmarlow:
@tsubery, laziness is the answer. I haven't seen many uses of those for docs. I wrote a few more tests and it was a problem with those cases as well, which leads to #235.

PR #234, opened 2016-12-02
Fix doc string termination when defined with sigil

tsubery:
This fixes more highlight termination cases that were not covered in #232 fix. For example:
image


tsubery:
Duplicate of #235

PR #240, opened 2016-12-19
Fix heredoc syntax highlighting

### LINKED ISSUE #239: Broken highlighting in heredocs in tests
First off, thanks for this project! Makes using Vim & Elixir a lot easier 😄 

I noticed recently that highlighting is broken for heredocs, but only in my tests for some reason. For example, I was working on Credo and I see this in a test file:

highlighting

It looks like it's not seeing the closing `"""` of the heredoc for some reason. I've tried assigning the heredoc to a variable so the closing `"""` could be on its own line, but that didn't solve the issue. I also tried changing the file extension to `.ex` from `.exs`, and that didn't do anything, either. I saw #230 and just updated all my plugins (I'm using Vundle), and still no dice.

Let me know if there's any more info or debugging data I can pass along to help fix this or if there's something I'm just totally missing.

Haegin:
Fixes #239.

Heredocs with a closing ''' or """ at the start of the line (with no preceeding
spaces) were being skipped because of the skip rule. Removing the skip rules
fixes the bug.

I wasn't able to get the tests to run on my machine as it seems to be looking for gVim which I don't have on OS X and changing it to just start Vim resulted in it not being able to find Vim. I also tried running them on an Ubuntu VM (again, without gVim as it's headless) but it seems to still need an X server. I think they should pass with the change (and fail without the fix) but it'd be good to check if someone has the test environment set up correctly.

### DIFF
--- a/spec/syntax/strings_spec.rb
+++ b/spec/syntax/strings_spec.rb
@@ -55,5 +55,21 @@ describe 'String syntax' do
       end
       EOF
     end
+
+    it 'correctly terminates heredocs with no spaces at the start of the line' do
+      expect(<<~'EOF'.gsub(/^\s+/, '')).to include_elixir_syntax('elixirAtom', ':bar')
+      """
+      foo
+      """
+      :bar
+      EOF
+
+      expect(<<~'EOF'.gsub(/^\s+/, '')).to include_elixir_syntax('elixirAtom', ':bar')
+      '''
+      foo
+      '''
+      :bar
+      EOF
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


jbodah:
Tests pass for me and did a quick manual verification

Haegin:
Is there anything more you need from me to get this merged down or is it just waiting on a review/merge/deploy?

jbodah:
cc: @kassio

Haegin:
Anything I can do to help get this out?

Issue #237, opened 2016-12-09
indent ecto keyword queries


jbodah:
I think we should indent Ecto queries as in the following example from the docs: https://hexdocs.pm/ecto/Ecto.Query.html#having/3-keywords-example

Currently this expected indentation:

```ex
defmodule New do
  def do_query do
    from user in Users,
      select: user.name,
      join: signup in Signups, where: user.id == signup.user_id
  end
end
```

Will indent as:

```ex
defmodule New do
  def do_query do
    from user in Users,
    select: user.name,
    join: signup in Signups, where: user.id == signup.user_id
  end
end
```

kassio:
I don't know if I agree, since the `from` is a function, and I think all the parameters of a function must be indented one extra level if they are in the following likes. The documentation itself has some examples that follow what I'm saying:

<img width="1194" alt="ecto_query_ _ecto_v2_1_1" src="https://cloud.githubusercontent.com/assets/120483/21651719/a49938fa-d2a1-11e6-89ca-ab98c691a160.png">
> https://hexdocs.pm/ecto/Ecto.Query.html

jbodah:
I think we're saying the same thing, right? I may have trickily switched up the actual/expected ordering here

PR #241, opened 2016-12-22
Fix faulty indentation of Ecto queries

DanielZendejas:
Hi,

This PR addresses [this issue](https://github.com/elixir-lang/vim-elixir/issues/237) and also cleans up a couple of lines of documentation I found.

This is my first experience ever with Vimscript, so please let me know what you think. Any comment or suggestion is welcome.


pbogut:
I didn't check it but how it will work if I have variable named `from`? Something like:
```elixir
def smth do
  from = 1
  to = 7
.....
end
```

DanielZendejas:
@pbogut good point, I did not think of that. I'll rewrite this later today so it doesn't indent wrongly.

DanielZendejas:
I fixed the `from` pattern and added a test based on the comment made by @pbogut . I'm not sure why the build failed though, the tests run fine in my local machine. 64 tests failed but I don't think it was because of these changes... did I do something wrong?

Anyway, the code is ready for another review, please let me know what do you think and if I'm missing something.

autoload/elixir/indent.vim
```diff
@@ -14,6 +14,7 @@ let s:MULTILINE_BLOCK = '\%(\<do\>'.s:NO_COLON_AFTER.'\|'.s:MULTILINE_FN.'\)'
 let s:BLOCK_MIDDLE = '\<\%(else\|match\|elsif\|catch\|after\|rescue\)\>'
 let s:BLOCK_END = 'end'
 let s:STARTS_WITH_PIPELINE = '^\s*|>.*$'
+let s:QUERY_FROM = '^\s*from\s*.*\s*in\s*.*,'
```
kassio: I think we need to improve this regexp, today it matchs with undesired texts too:
nvim

I'd use something like: `^\s*\<from\>.*\<in\>.*,`, wdyt?

DanielZendejas: You're right, I'll change it later today.

DanielZendejas:
I've replaced the regexp with the one suggested by @kassio and added the example provided as a test.

PR #244, opened 2016-12-23
add syntax for tuples, structs, and maps

### LINKED ISSUE #242: highlight: brace mismatch when interpolated expression contains tuple
In the following code the trailing `}` in `{:unexpected_status_code, s}` is incorrectly highlighted:

```ex
"Failed sending tasks #{inspect {:unexpected_status_code, s}}"
```

Screenshot:

1__hello_ex______repos_strap__-_vim__vim_

jbodah:
* fixes #242
* adds tests for interpolated data structures

### DIFF
--- a/spec/syntax/map_spec.rb
+++ b/spec/syntax/map_spec.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'Map syntax' do
+  it 'maps' do
+    str = %q(%{name: "josh"})
+    expect(str).to include_elixir_syntax('elixirAtom', 'name:')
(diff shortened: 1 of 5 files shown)


remiprev:
@jbodah I have a small issue with this pull request. This code

```elixir
%Foo{bar: true}
```

now considers `%Foo{` as the whole `elixirStructDelimiter`. I liked it when `%` and `{` were a different color than `Foo` and I’d like to revert this locally 😄.

I’m not familiar at all with `syn region` but I was wondering if there is a quick fix that I could add to this line to make it highlight `Foo` as a `elixirAlias` like before this PR.

Thank you!

jbodah:
@remiprev nice catch. We should be able to switch the regex to use look behind. I'm currently away but I can take a peek in a day or two

PR #256, opened 2017-01-11
indent: clear pipeline indentation after blank new line

### LINKED ISSUE #227: Incorrect indentation for fn when used with pipe
Thanks for creating this plugin :+1: 

There are instances of using `fn` with pipes which causes incorrect indentations. Here is the minimal code to reproduce the issue.

### Expected 1

```
names = ~w(Mujju Zainab)
upcased_names = names
                |> Enum.map(fn name ->
                  String.upcase(name)
                end)

IO.inspect names
```

### Actual 1

```
names = ~w(Mujju Zainab)
upcased_names = names
                |> Enum.map(fn name ->
                  String.upcase(name)
                end)

                IO.inspect names
                #^^^^^^^^^^^^^^^^
                #incorrect indentation
```

### Expected 2
```
names = ~w(Mujju Zainab)
upcased_names = names
                |> Enum.map(fn name ->
                  String.upcase(name) end)

IO.inspect names
```

### Actual 2
```
names = ~w(Mujju Zainab)
upcased_names = names
                |> Enum.map(fn name ->
                  String.upcase(name) end)

                  IO.inspect names
                  #^^^^^^^^^^^^^^^^
                  #incorrect indentation
```

### Expected == Actual
```
names = ~w(Mujju Zainab)
upcased_names = names
                |> Enum.map(fn name -> String.upcase(name) end)

IO.inspect names
```

jbodah:
Don't have a ton of time to clean this up at the moment, but here's the
result of a quick hack session. I think something like
`line.last`, `line.last_non_blank`, and `line.current` might make more
sense

* fixes #227

### DIFF
--- a/spec/indent/pipeline_spec.rb
+++ b/spec/indent/pipeline_spec.rb
@@ -119,4 +119,36 @@ describe 'Indenting pipeline' do
       EOF
     end
   end
+
+  it 'resets the indent after a blank new line' do
+    expect(<<~EOF).to be_elixir_indentation
+      upcased_names = names
+                      |> Enum.map(fn name ->
+                        String.upcase(name)
+                      end)
+
+      IO.inspect names
+    EOF
+  end
+
+  it 'resets the indent after a blank line pt. 2' do
+    expect(<<~EOF).to be_elixir_indentation
+      upcased_names = names
+                      |> Enum.map(fn name ->
+                        String.upcase(name) end)
+
+      IO.inspect names
+    EOF
+  end
+
+  it 'keeps indent after a blank if current starts with pipe' do
+    expect(<<~EOF).to be_elixir_indentation
+      upcased_names = names
+                      |> Enum.map(fn name ->
+                        String.upcase(name)
+                      end)
+
+                      |> do_stuff
+    EOF
+  end
 end
--- a/autoload/elixir/indent.vim
+++ b/autoload/elixir/indent.vim
@@ -25,20 +25,20 @@ let s:LINE_COMMENT = '^\s*#'
 let s:MATCH_OPERATOR = '[^!><=]=[^~=>]'
 
 function! s:pending_parenthesis(line)
-  if a:line.last.text !~ s:ARROW
-    return elixir#util#count_indentable_symbol_diff(a:line.last, '(', '\%(end\s*\)\@<!)')
+  if a:line.last_non_blank.text !~ s:ARROW
+    return elixir#util#count_indentable_symbol_diff(a:line.last_non_blank, '(', '\%(end\s*\)\@<!)')
   end
 endfunction
 
 function! s:pending_square_brackets(line)
-  if a:line.last.text !~ s:ARROW
-    return elixir#util#count_indentable_symbol_diff(a:line.last, '[', ']')
+  if a:line.last_non_blank.text !~ s:ARROW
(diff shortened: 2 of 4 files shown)


kassio:
First, you can rename things easily on all the project running (if you already have `ag`):
```console
ag -l --nocolor "line.last" $* | xargs sed -i "" -e "s/line.last/line.last_non_blank/g"
```
btw, this `replace` is so normal on my life that I created a [`replace` command](https://github.com/kassio/dotfiles/blob/master/bin/replace) on my env.

Also, I liked the idea of `non_blank`, what do you think to use a more power data structure for the lines:
```viml
function! s:build_line(line)
  let line = { 'current': {}, 'last': {}, 'last_non_blank': {} }
  let line.current = s:new_line(a:line)
  let line.last = s:new_line(line.current.num - 1)
  let line.last_non_blank = s:new_line(prevnonblank(line.current.num - 1))

  return line
endfunction

function! s:new_line(num)
  let line = {
        \ "num": a:num,
        \ "text": getline(a:num)
        \ }

  function line.is_blank()
    return self.text =~ '\s*$'
  endfunction

  return line
endfunction
```
this way we can do:
```viml
" ...
elseif exits("b:old_ind.pipeline") && a:line.last.is_blank()
" ...
```

jbodah:
I like it, my only feedback would be `line.last.is_blank()` vs. `is_blank(line.last)` with my preference being the latter if we're nitpicking - both sound great though, and I like the new data structure's format.

Random side note - do you happen to know how to log the data structure or which data structures vim supports (e.g. is there a spot in the help docs)? I've had a painful time trying to do something like `:echom line` or any other type of puts debugging

kassio:
Let me check if I got it right, do you rather use a global function? something like `elixir#util#is_blank(string)` passing the line? 

To print structures with `echom` you can use the `string()` function, like this: `:echom string({"a": 1})`

jbodah:
> Let me check if I got it right, do you rather use a global function? something like elixir#util#is_blank(string) passing the line?

Eh idk I'm pretty indifferent - feel free to decide and I'll fix the conflicts/make the changes

kassio:
Ok, so lets use a global function in `elixir#util` namespace:
```viml
" vim-elixir/indent/elixir.vim
function! s:build_line(line)
  let line = { 'current': {}, 'last': {}, 'last_non_blank': {} }
  let line.current = s:new_line(a:line)
  let line.last = s:new_line(line.current.num - 1)
  let line.last_non_blank = s:new_line(prevnonblank(line.current.num - 1))

  return line
endfunction

function! s:new_line(num)
  return {
        \ "num": a:num,
        \ "text": getline(a:num)
        \ }
endfunction

" vim-elixir/autoload/elixir/util.vim
function elixir#util#is_blank(string)
  return a:string =~ '^\s*$'
endfunction
```

kassio:
news on this one?

spec/indent/pipeline_spec.rb
```diff
@@ -119,4 +119,36 @@ def build_command(input, output) do
...
+    expect(<<~EOF).to be_elixir_indentation
+      upcased_names = names
+                      |> Enum.map(fn name ->
+                        String.upcase(name) end)
+
+      IO.inspect names
+    EOF
+  end
+
+  it 'keeps indent after a blank if current starts with pipe' do
```
jbodah: Added this test case

jbodah:
@kassio should be good now

kassio:
Great job!

PR #269, opened 2017-01-23
indent: use regex to correctly determine `end` keyword

### LINKED ISSUE #267: Indentation problem with line including word 'end'
When selecting this text and pressing `=` it gets indented wrongly

```elixir
  fun1 = fn
    (:foo) ->
      :bar
      :end
  end

  fun2 = fn
    (:foo) ->
      :bar
      'end'
  end

  fun3 = fn
    (:foo) ->
      :bar
      :send
  end
```

## Actual

```elixir
  fun1 = fn
    (:foo) ->
      :bar
    :end
  end

  fun2 = fn
    (:foo) ->
      :bar
    'end'
end

fun3 = fn
  (:foo) ->
    :bar
  :send
end
```

Pressing `<Enter>` after `:bar` puts cursor in correct position. It is just when whole text selected and `=` is pressed problem occurs.

jbodah:
* fixes #267


gasparch:
Btw does not fix this case

```elixir
  fun2 = fn
    (:foo) ->
      :foo
    "string with end"
    end
```

We really need to look into syntax attributes to determine if it is a keyword.

jbodah:
You're right, this needs to go through `elixir#util#is_indentable_at`

autoload/elixir/indent.vim
```diff
@@ -62,6 +62,14 @@ function! elixir#indent#deindent_ending_symbols(ind, line)
...
+  echom "> pair start ". s:PAIR_START
+  echom "> pair middle ". s:PAIR_MIDDLE
+  echom "> pair end ". s:PAIR_END
+  echom "> pair skip ". s:PAIR_SKIP
+
+  " NOTE: cursor is not being properly set here which causes this to
+  " incorrectly handle nesting
+  echom "> cursor => " . line(".") ", " . col(".")
   if a:line.current.text =~ s:DEINDENT_KEYWORDS
     let bslnum = searchpair(
```
jbodah: @kassio @gasparch not sure if you have any ideas around this: 340b281ad6b38513337465337677851c407d608d fixes this issue however the `searchpair` call is not always correct.

It looks like the trailing `\zs` in `PAIR_END` causes the following tests to still fail. Any thoughts why this is the case? Removing it causes several other tests to fail

```
Failures:

  1) Indenting anonymous functions fn with comment with "end"
     Failure/Error:
       expect(<<~EOF).to be_elixir_indentation
       fun3 = fn
         (:foo) ->
           :foo
           :asd # end
       end
       EOF

       @@ -2,5 +2,5 @@
          (:foo) ->
            :foo
            :asd # end
       -end
       +  end
     # ./spec/indent/anonymous_functions_spec.rb:111:in `block (2 levels) in <top (required)>'

  2) Indenting anonymous functions fn with string with "end"
     Failure/Error:
       expect(<<~EOF).to be_elixir_indentation
       fun2 = fn
         (:foo) ->
           :foo
           "string with end"
       end
       EOF

       @@ -2,5 +2,5 @@
          (:foo) ->
            :foo
            "string with end"
       -end
       +  end
     # ./spec/indent/anonymous_functions_spec.rb:101:in `block (2 levels) in <top (required)>'

  3) Indenting documentation with end keyword
     Failure/Error:
       expect(<<~EOF).to be_elixir_indentation
       defmodule Test do
         @doc """
         end
         """
       end
       EOF

       @@ -2,5 +2,5 @@
          @doc """
          end
          """
       -end
       +  end
     # ./spec/indent/documentation_spec.rb:7:in `block (2 levels) in <top (required)>'

Finished in 39.72 seconds (files took 0.12091 seconds to load)
200 examples, 3 failures, 5 pending

Failed examples:

rspec ./spec/indent/anonymous_functions_spec.rb:110 # Indenting anonymous functions fn with comment with "end"
rspec ./spec/indent/anonymous_functions_spec.rb:100 # Indenting anonymous functions fn with string with "end"
rspec ./spec/indent/documentation_spec.rb:6 # Indenting documentation with end keyword
```

kassio: @jbodah I was trying to understand why the `searchpair` is not working, and my findings are:
1. the `\zs` at the end of the `s:PAIR_END` is what make these tests break, however, if we remove it the other tests will break.
2. Also find another case where the indentation will fail:
```elixir
(:foo) ->
  :foo # fn
  :asd
end
```

With that in mind I think that we'll need to make this function check more details:
1. Instead of `if a:line.current.text =~ s:DEINDENT_KEYWORDS` we'll need something like `if has_indentable_match(a:line.current, s:DEINDENT_KEYWORDS)`
2. Instead of `searchpair` I think we'll need to use the `searchpairpos` and check if the given `pos` is indentable
3. Understand better the reason of the `\zs` (I don't remember, I just know that it's related to indent the `end` word while typing.)

jbodah:
Closing in favor of #289

PR #270, opened 2017-01-23
indent: skip strings/comments when checking for open-brace indentation

### LINKED ISSUE #266: Incorrect indentation w/curly brackets inside strings
## Expected 

```elixir
  x = [
    {:text, "asd {"},
    {:text, "qwe"},
  ]
```

## Actual

```elixir
  x = [
    {:text, "asd {"},
      {:text, "qwe"},

    ]
```

Problem does not occur with `[` or `(` . Problem exists in HEAD right now (3dbcec)

jbodah:
* fixes #266

### DIFF
--- a/autoload/elixir/util.vim
+++ b/autoload/elixir/util.vim
@@ -26,23 +26,25 @@ function! elixir#util#count_indentable_symbol_diff(line, open, close)
   if elixir#util#is_indentable_match(a:line, a:open)
         \ && elixir#util#is_indentable_match(a:line, a:close)
     return
-          \   s:match_count(a:line.text, a:open)
-          \ - s:match_count(a:line.text, a:close)
+          \   s:match_count(a:line, a:open)
+          \ - s:match_count(a:line, a:close)
   else
     return 0
   end
 endfunction
 
-function! s:match_count(string, pattern)
-  let size = strlen(a:string)
+function! s:match_count(line, pattern)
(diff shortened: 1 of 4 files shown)


kassio:
Some tests are breaking with this:
```console
rspec ./spec/indent/case_spec.rb:47 # Indenting case statements type case..do..end
rspec ./spec/indent/case_spec.rb:30 # Indenting case statements nested case statements
```

autoload/elixir/util.vim
```diff
@@ -26,23 +26,25 @@ function! elixir#util#count_indentable_symbol_diff(line, open, close)
...
   let index = 0
   let counter = 0
 
   while index < size
-    let index = match(a:string, a:pattern, index)
+    let index = match(a:line.text, a:pattern, index)
     if index >= 0
       let index += 1
-      let counter +=1
+      if elixir#util#is_indentable_at(a:line.num, index)
```
kassio: I think the `let index += 1` must be done after this `if`, right?

kassio:
These broken tests on CI seems to be a CI problem, just move the `let index += 1`, and I'll merge it.

autoload/elixir/util.vim
kassio: Actually, the `is_indentable_at` uses the `column` and not the index. So your code is right in increment the index before checking. However, this means that the `is_indentable_match` function has a bug :(

PR #278, opened 2017-02-04
changed colors of modules and aliases

iansowinski:
Now module names have the same color as keywords. Syntax coloring also does't work as it should when it comes to coloring what's after 'require' keyword. This is affected by elixirAlias. That's my proposal of changes.

Before:
zrzut ekranu 2017-02-04 14 21 31

After:
zrzut ekranu 2017-02-04 14 51 44

### DIFF
--- a/syntax/elixir.vim
+++ b/syntax/elixir.vim
@@ -189,7 +189,7 @@ hi def link elixirExceptionDefine        Define
 hi def link elixirCallbackDefine         Define
 hi def link elixirStructDefine           Define
 hi def link elixirExUnitMacro            Define
-hi def link elixirModuleDeclaration      Type
+hi def link elixirModuleDeclaration      Identifier
 hi def link elixirFunctionDeclaration    Function
 hi def link elixirMacroDeclaration       Macro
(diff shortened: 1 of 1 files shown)


iansowinski:
I don't have this kind of issues wyth vim-ruby and on different color schemes (I use [vim-monokai](https://github.com/sickill/vim-monokai), but on default vim colors it doesn't work as well)

jbodah:
Are you on HEAD? If so, could you please go through the following steps:

Add this snippet to your vimrc

```vim
map <leader>syn :echo "hi<" . synIDattr(synID(line("."),col("."),1),"name") . '> trans<'
      \ . synIDattr(synID(line("."),col("."),0),"name") . "> lo<"
      \ . synIDattr(synIDtrans(synID(line("."),col("."),1)),"name") . ">"<CR>
```

Open the file above. Put the cursor over `ModuleOne` and type `<leader>syn`. Please show me the output you get

iansowinski:
It seams, that you're right. I fixed my color scheme instead and everythinig works as it should. Thanks for help, I'm closing PR.

PR #280, opened 2017-02-11
fixed typo in 'indentkeys'

nhooyr:


jbodah:
@nhooyr did this break under a specific case? Based on the help docs for `indetkeys-format` I believe this was intended

nhooyr:
Hmm, you are correct. Indentation was not functioning correctly for me on hitting `<CR>` until I made this change. I'll look into it further and let you know what I find.

jbodah:
A few things to check:

Are you on HEAD? Does it still happen with other plugins disabled?

If so, an example would be helpful. I'm currently doing a large pass at the indent code

nhooyr:
I can't seem to reproduce. Must have been user error, sorry for the noise

jbodah:
No worries 👍

PR #289, opened 2017-02-27
Indent rewrite

jbodah:
This PR looks to simplify the indent code as well as make it stateless. Currently all of the existing tests pass as well as some I added while spiking.

TODO
- [x] Refactor new indent code
- [x] Clean up old indent files
- [x] Investigate other little differences between old indent code and new code
- [x] Check all open indent issues with the new code


kassio:
As I see you moved back all the function to one file. I'm against this idea for two, main, reasons:
1. IMHO it's hard to maintain large amount of code in a single file.
2. The functions that are defined in the `autoload` folder are loaded only when they are needed, making the vim initialisation process faster (`:help autoload-functions`): 

> When using many or large functions, it's possible to automatically define them
only when they are used.  There are two methods: with an autocommand and with
the "autoload" directory in 'runtimepath'.

PS.: I'm not trying to protect the old code, tbh, I didn't analise the new code. I just, strongly, recommend you to keep most of the functions within the `autoload` folder.

jbodah:
I'm cool with moving back to autoload and breaking this into functions again. The main complexities I want to move away from are buffer state and cascading (if possible I think branching will be easier to maintain. My guess is we'll need both at some point)

kassio:
@jbodah yeah, I tried to remove the buffer state when I rewrote the code, however, I wasn't able to do so, I hope you find a solution. And, I repeat, I'll be glad if you need any help. :)

jbodah:
Most of it is covered by tests which has been super helpful. I do have some questions about some of the config though:

```
setlocal nosmartindent
 
let s:cpo_save = &cpo
set cpo&vim
```

I haven't been able to figure out what these two lines are for. I noticed some other indent libraries ( ([vim-ruby](https://github.com/vim-ruby/vim-ruby/blob/master/indent/ruby.vim) and [vim-go](https://github.com/fatih/vim-go/blob/master/indent/go.vim)) doing similar things but I haven't been able to come up with a case where *not* setting these causes an issue

kassio:
tbh, I'm not sure *why* these settings are there. What I know is, this plugin got a lot of its initial settings from vim-ruby, as you have figured out. I think these avoid some not planned behaviour, since most of the vim default indent settings work, mostly, for c-like languages.

jacknoble:
This is great -- so great I'm temporarily pointing vim toward your branch. Since it fixes so many bugs do you think it would be better to finish up the PR (after moving functions back to autoload) and address your other TODOs in separate pull requests?

jbodah:
> Since it fixes so many bugs do you think it would be better to finish up the PR (after moving functions back to autoload) and address your other TODOs in separate pull requests?

I think that makes sense. I've been using this branch locally for the past week or so and it's been reasonable (still a few edge cases here and there). I'll clean this up over the next day or so and see if we can get it merged

jbodah:
@kassio what do you think of something like this? Basically each handler is a branch and only a single handler is ever used which should simplify debugging an individual indent case

There are probably some optimizations to more efficiently find which case we're in, and there are cases that could be condensed into a single case, but this has all the tests passing and fixes several issues in the backlog

jbodah:
Oh wait, I need to revert all of the `+ 2` hacks and use the actual tab width

jbodah:
Going to merge this - happy to address feedback retrospectively

PR #337, opened 2017-05-18
Performance optimizations

jbodah:


jbodah:
Things to try:

- in `handle_inside_nested_construct` match on larger patterns (e.g. `receive do`)
- in `handle_inside_nested_construct` match with lookbehind (e.g. `try` without an non-`=` non-whitespace characters preceding it in the line)

PR #345, opened 2017-06-27
Refactoring indent + add max indent lookbehind

jbodah:
This PR cleans up the indent code and adds a configurable max indent lookbehind value which can be used to stop `searchpair` from killing performance. The default value is 30 which feels like it should be good enough for most clean, idiomatic elixir code. We can try and improve this with a search for the nearest scope bound (e.g. `defmodule`, `def`) in the future for a better experience

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+profile.log
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 profile.log
+test_indent.result
--- a/Gemfile
+++ b/Gemfile
@@ -2,3 +2,5 @@ source 'https://rubygems.org'
 
 gem 'rspec'
 gem 'vimrunner'
+gem 'pry'
+gem 'diffy'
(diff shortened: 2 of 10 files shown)


gasparch:
From a first glance seems to work quite good for me. Rensposivness seems to be okay, even with max_look_back = 100.

Big is my file with 90kb size, small is one supplied in repo (30k).

[profile-big.txt](https://github.com/elixir-lang/vim-elixir/files/1106587/profile-big.txt)
[profile-small.txt](https://github.com/elixir-lang/vim-elixir/files/1106588/profile-small.txt)

I'll report if there are some breakage of indent logic.

Still I would think of adding cache/memoization on top of it, just to speed up even more :)

gasparch:
@jbodah ping? :)

jbodah:
@gasparch I've been using this for about 2 weeks or so without too much trouble so I'm going to merge it when I get back into the office. I will likely rollback c31962f and open a new branch for that

gasparch:
super cool. so far it was good for me too.

PR #355, opened 2017-07-13
spec_helper: detect failed searches

mhinz:
`@vim.search()` silently fails if a pattern is not found and the cursor won't move. So, if the current cursor position happens to sport the expected syntax group already, this can lead to false positive tests.

We work around this by using Vim's `search()` function, which returns 0 if there is no match.


jbodah:
Awesome, you're killing it :rocket:

Issue #340, opened 2017-06-19
Case statement has invalid highlight


jbodah:
```ex
  def get_success_rate do
    %{success_rate: success_rate, completed: completed, failed: failed} = Task.Stats.SQL.compute(10, :minute)

    case {completed, failed} do
      {nil, nil} -> 1
      {nil, failed} when failed > @min_count                    -> parse_success_rate(success_rate)
      {completed, nil} when completed > @min_count              -> parse_success_rate(success_rate)
      {completed, failed} when completed + failed > @min_count  -> parse_success_rate(success_rate)
      _ -> 1
    end
  end
```

mhinz:
What's wrong with it?

jbodah:
@mhinz I believe it had to do with using guards in the match clauses. Not near my machine to test it though

mhinz:
Ah, the guard breaks the highlighting of `end`.

When you query the highlight groups from the last `end`, you'll get:

```
['elixirBlock', 'elixirGuard', 'elixirGuard', 'elixirTuple', 'elixirBlock', 'elixirGuard
', 'elixirGuard', 'elixirGuard', 'elixirGuard', 'elixirTuple', 'elixirGuard', 'elixirGua
rd', 'elixirGuard', 'elixirGuard', 'elixirTuple', 'elixirId']
```

That's madness! I'm looking into it.

PR #357, opened 2017-07-17
syntax: fix highlighting of Kernel functions

mhinz:
Prior to this change, there was a false assumption that Kernel functions can only be used in guards.

But they can appear everywhere where an identifer can appear. Following this rule we can greatly simplify the highlighting of Kernel functions.

Fixes https://github.com/elixir-lang/vim-elixir/issues/340

### DIFF
--- a/spec/syntax/guard_spec.rb
+++ b/spec/syntax/guard_spec.rb
@@ -5,24 +5,30 @@ require 'spec_helper'
 describe 'Guard syntax' do
   it 'guard in function' do
     expect(<<~EOF).to include_elixir_syntax('elixirKernelFunction', 'is_atom')
-    def fun(a) when is_atom(a) do
-    end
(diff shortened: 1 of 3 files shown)


jbodah:
LGTM - much simpler. Verified the tests locally too. Thanks for looking into this

mhinz:
Ah, there's just one thing to fix which is not covered by tests yet:

```elixir
plug Plug.Parsers, length: 400_000_000
```

Here, `length` is highlighted as `elixirKernelFunction` although it should be `elixirAtom`.

Only this is tested:

```elixir
do
  plug Plug.Parsers, length: 400_000_000
```

mhinz:
Okay, I simply made the keywords contained now and added a test for the code mentioned in the last comment.

Should be fine now!

jbodah:
:shipit:

PR #358, opened 2017-07-17
indent: add follow_prev_nb

### LINKED ISSUE #353: Deep block-level indent fails following @spec
# Actual

```ex
defmodule MyModule do
  # 30 blank lines
  @spec something :: something
def my_fun do

end
end
```

The problem has to do with how we handle defaulting. Do we follow the prev non-blank line or do we try and trace it up (e.g. following nesting due to binary operator chaining/assignment/pipelines)? Right now there is a hack to handle `end` properly, but nothing for any module attributes. Still - I think we should handle this differently than whitelisting edge cases until tests pass

jbodah:
Adds better rules for following the indent of the previous nonblank line. Fleshes out a quick hack that only followed lines that end with `end`. Fixes #353

### DIFF
--- a/autoload/elixir/indent.vim
+++ b/autoload/elixir/indent.vim
@@ -23,7 +23,7 @@ function! elixir#indent#indent(lnum)
         \'inside_nested_construct',
         \'starts_with_comment',
         \'inside_generic_block',
-        \'following_prev_end'
+        \'follow_prev_nb'
         \]
   for handler in handlers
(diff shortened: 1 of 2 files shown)


jbodah:
Specs pass despite what CI says; working off this branch for a bit and will merge if everything is :100:

mhinz:
I haven't found time to read up on the indent code yet, but.. LGTM. :-)

PR #359, opened 2017-07-19
syntax: highlight "into:" as inline block

### LINKED ISSUE #312: syntax: func references screw up trailing keyword highlight
`into` highlight is incorrect

```ex
  defp atomize_keys(map), do: for kvp <- map, do: &atomize_key/1, into: %{}
```

mhinz:
Fixes #312

### DIFF
--- a/spec/syntax/atom_spec.rb
+++ b/spec/syntax/atom_spec.rb
@@ -41,4 +41,10 @@ describe 'Atom syntax' do
       EOF
     end
   end
+
+  it 'atoms as part of a comprehension' do
+    s = 'for kvp <- map, do: &atomize_key/1, into: %{}'
+    expect(s).to include_elixir_syntax('elixirAtom', 'do')
+    expect(s).to include_elixir_syntax('elixirAtom', 'into')
+  end
 end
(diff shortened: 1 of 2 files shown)


jbodah:
LGTM - `elixirBlockInline` seems like the wrong approach to me in the long-term however. Macros can accept any keyword arguments, so there is an unbounded set of things we'd have to account for. I think the right long-term approach would be to lump these into `elixirAtom`s

mhinz:
Let's fix it right away then!

The latest commit simply removes the whole `elixirBlockInline` syntax group which makes the code to fall back to `elixirAtom` instead.

jbodah:
One last thing - would you mind adding a test case for the syntax of `into:` in case of regressions when refactoring? Otherwise this looks 💯

mhinz:
Like this?

BTW, another thing that comes to mind.. in the case of `foo: bar`, the `:` also belongs to `elixirAtom`. Is that what we want?

jbodah:
> BTW, another thing that comes to mind.. in the case of foo: bar, the : also belongs to elixirAtom. Is that what we want?

I'm not sure tbh. I think yes but I don't feel strongly right now and would be happy with going finer grain if we had people interested in that, but it hasn't come up yet through any requests AFAIK

mhinz:
Does that mean _yes_ or _no_ for the colon question? :)

jbodah:
For what it's worth, redcarpet seems to do this too ^

mhinz:
Yeah, I like this change.

It makes sense to highlight `do:`, used in an actual atom context, differently from the `do` macro that is used for structural purposes.

PR #360, opened 2017-07-19
spec_helper: match syntax group exactly

mhinz:
Previously, syntax() would return a string containing a Vim list:

```
  "['elixirBlockDefinition']".include? 'elixirBlock'
```

This would match although it is not what we want.

This patch converts the return value of syntax() from a string to a list:

```
  ["elixirBlockDefinition"].include? 'elixirBlock'
```

This does not match anymore.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -52,10 +52,13 @@ class Buffer
       return []
     end
 
-    # Return the syntax groups in a list.
-    @vim.echo <<~EOF
+    syngroups = @vim.echo <<~EOF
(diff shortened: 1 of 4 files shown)


mhinz:
And once again this found 5 false positive tests. :D 

Fixing them later.

jbodah:
Nice catch - LGTM. It looks like this breaks several test cases. Since this is test-only if you'd prefer to mark the failing tests as pending in order to get this in quicker then that'd be fine by me as well

jbodah:
🚀 🎸 💯

Issue #299, opened 2017-03-28
syntax: :queue.in should not be highlighted as a keyword


jbodah:
`hi<airline_x_to_airline_y_bold> trans<airline_x_to_airline_y_bold> lo<Statement>`

mhinz:
Sorry, but I'm confused. Has the example anything to do with the `:queue.in/2` mentioned in the title? :->

Anyway, I noticed this:

```elixir
x = :queue.new
x = :queue.in(5, x)
```

In this case, the `new` is highlighted as `elixirId`, whereas the `in` is highlighted as `elixirOperator`. (I guess that's what you meant.)

jbodah:
Yep - the problem is that the leading `.` in `.in` should cause it to be recognized not as an operator, but as an identifier/function name

mhinz:
Woah, this is tricky to get right.

E.g., `<-` is only highlighted as operator, because it matches twice! For `<` and for `-`. This means, just prepending something to the `elixirOperator` regexp to not match after `.` won't work, because in the case of `.<-`, the `<` won't be highlighted, but the `-` still will.

jbodah:
> Has the example anything to do with the :queue.in/2 mentioned in the title?

Sorry I just realized I never responded to this. No, the "example" isn't really an example and is frankly useless (it was the syntax groups but those are obscured by airline). The example you gave is what to use

> E.g., <- is only highlighted as operator, because it matches twice! For < and for -. This means, just prepending something to the elixirOperator regexp to not match after . won't work, because in the case of .<-, the < won't be highlighted, but the - still will.

Yeah - these are pretty fringe, but here's an example from the Kernel module: https://github.com/elixir-lang/elixir/blob/v1.4.5/lib/elixir/lib/kernel.ex#L1013

```ex
  def left ++ right do
    :erlang.++(left, right)
  end
```

Thus,  I think the `++` in the signature should be considered an operator, but the `.++` in the body is a function name. This is really fringe and almost no one calls these operators with a `.` so I'm indifferent on if/when we choose to support these, however the keyword "operators" (`and`, `or`, etc) should definitely be supported in case other modules use them (as is the case in `:queue.in`)

mhinz:
I agree. I'm playing around with different approaches right now, but when everything fails, I'll just change from..

```vim
syn keyword elixirOperator and not or in
```

to

```vim
syn match elixirOperator '\v\.@<!%(and|or|in|not)'
```

Then we at least catch the most common cases.

[...]

PR #361, opened 2017-07-20
syntax: highlight operators after a . as elixirId

mhinz:
Fixes https://github.com/elixir-lang/vim-elixir/issues/299

---

@jbodah What do you think about this approach?

It's easier than trying to make all the regexp of `elixirOperator` super smart.

Basically it's.. magic. After a `.` the regexp engine gets a hint to check for `elixirId` first. `elixirId` is defined by 2 patterns. The first matches the same as before and the new one only matches after a `.`, only graphical characters (`&`, `<` etc.), for up to 3 characters (the widest possible operators are `~~~`, `>>>`, etc.) or until the first space or `(`.

Is this sane? ;x

### DIFF
--- a/spec/syntax/operator_spec.rb
+++ b/spec/syntax/operator_spec.rb
@@ -2,7 +2,7 @@
 
 require 'spec_helper'
 
-describe 'Default argument syntax' do
+describe 'Operators' do
   it 'default argument' do
     expect(<<~'EOF').to include_elixir_syntax('elixirOperator', '\\')
       def foo(bar \\ :baz)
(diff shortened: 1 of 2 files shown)


mhinz:
I haven't even looked at the tests yet, but it works correctly for these cases:

```elixir
x = :queue.new
_ = :queue.in x, 5

Kernel.++([1], [2])
Kernel.++ [1], [2]
Kernel     . ++  [1]  , [2]
```

mhinz:
Okay, I decided to go for the simple solution which covers the most important cases, since the pattern is much easier to maintain for users and to compute for machines.

Tests added.

jbodah:
Looks great 🎉

PR #362, opened 2017-07-28
Use the effective value of 'shiftwidth'

gcmt:
Among users that use tabs the `shiftwidth` option is usually set to `0`. According to the docs when `shiftwidth=0` the value of `tabstop` is used (see `:help 'shiftwidth'`). 

Vim already provides a function for retrieving the effective value of `shiftwidth`, `shiftwidth()`, but according to the docs it was introduced only in Vim 7.3 (see `:help shiftwidth()`).


jbodah:
Looks great - thanks!

PR #365, opened 2017-07-30
syntax: match and/or/in/not as operators only if they are entire words

### LINKED ISSUE #364: syntax: operators are incorrectly highlighted inside elixirId
Here's an example:

1__hello_ex______repos_vim-elixir__-_vim__vim_

Shout-out to @hauleth for catching this 👼

jbodah:
Fixes #364

### DIFF
--- a/spec/syntax/operator_spec.rb
+++ b/spec/syntax/operator_spec.rb
@@ -22,4 +22,10 @@ describe 'Operators' do
       :queue.in x, 5
     EOF
   end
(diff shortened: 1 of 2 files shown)


jbodah:
> Ouch. My bad.

No worries, I'm very much of the opinion "if the tests pass, :shipit:". It causes some occasional pain but makes our test suite stronger in the long run and lets us move faster

PR #371, opened 2017-09-14
Support interpolation in docstrings

glittershark:
Before:
before
After:
after


jbodah:
Looks great - would you mind adding a simple test to catch any regressions?

PR #372, opened 2017-09-18
Fix indentation issue with vim 8 - fixes #370

### LINKED ISSUE #370: Different autoindent behaviors between `else` and `end` inside of blocks
This is more of a question rather than an issue, so I apologize in advanced if this isn't the appropriate place for this.

I've noticed the auto-indenting behavior differs for `end` and `else` keywords when inside of a block. When typing `else` inside of an `if` block it will auto-indent to align with the opening `if` as soon as you finish typing the word. With `end` inside of a `do` block you have to press enter and insert a new line in order for the indentation to occur. `rescue` and `catch` also have the same behavior as `end` where a newline is required.

Is this expected behavior? I'm wondering if this is intentional to avoid frustration from certain ambiguities (something like `ending = get_ending()` triggering the indent), or possibly for performance reasons when dealing with nested `do` blocks.

I feel like the two behaviors were the same before I updated (not sure what exact commit I was on), though I can't seem to revert to commit where that is the case, so maybe there is something else going on here.

I prefer to have the indentation occur as soon as the word is finished typing, as my workflow involves using `alt+<key>` in order to exit insert mode and issue a command in the same keystroke (typically a directional command). Currently it requires entering a new line in insert mode to trigger the indent which leaves a superfluous blank line that I need to delete and doesn't allow me to exit insert mode as soon as I'm finished typing `end`.

Understandably everyone has different preferences and expectations, so if this behavior is intentional could someone point me in the direction of where this behavior is defined in the code so I can make the adjustment for myself? I've been learning a little bit about vim plugins and trying to familiarize myself with the the code, but I can't find where this behavior is defined.

# Actual

```ex
def foo() do
    :bar
    end▒▏
```

# Expected

```ex
def foo() do
    :bar
end▒▏
```

chrisdambrosio:


jbodah:
Seems reasonable to me. I'd like to get to the bottom of whether the regression was intentional, but I can't think of any problems with indenting more often. Thanks for the contribution

PR #373, opened 2017-09-18
update docs to use correct repo

gregors:
The current docs don't work. Updated repo location so that plugins actually get installed.


jbodah:
Good catch, thanks!

PR #377, opened 2017-10-11
syntax: support interpolation in docstrings

mhinz:
I picked up the commit from #371 and changed the test that explicitly checked for _no_ interpolation before.

Closes https://github.com/elixir-editors/vim-elixir/pull/371.


glittershark:
Hey, thanks! Sorry I never got back to that commit.

PR #391, opened 2017-12-31
Set define value

### LINKED ISSUE #387: Set values for include-search
Support for `'include'` would be tricky, but support for `'define'` should be easy. While Elixir doesn't have macros in Vim (aka C) sense it would be useful to use it as quick lookup for function/macro definitions via:

```
let &l:define = 'def\(macro\)\?p\?'
```

hauleth:
This allows using `:dlist` to list all functions defined in current
file.

Close #387

### DIFF
--- a/ftplugin/elixir.vim
+++ b/ftplugin/elixir.vim
@@ -1,10 +1,10 @@
-if (exists("b:did_ftplugin"))
+if exists('b:did_ftplugin')
   finish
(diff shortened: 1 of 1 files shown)


hauleth:
I am wandering wether it should also match `defmodule` or not.

jbodah:
LGTM thanks

PR #406, opened 2018-02-01
Add defguard(p) syntax highlighting

seddy:
I've tried this out locally and it appears to work fine, for example:

defguard_highlights

I didn't add any tests as there didn't appear to be any around syntax highlighting - happy to do so if I've just missed that :)


jbodah:
Thanks for this - this look great. We do test that the regex properly matches to the syntax group. Here's an example of one: https://github.com/elixir-editors/vim-elixir/blob/master/spec/syntax/exunit_spec.rb#L7

seddy:
Ah, great thanks @jbodah - I'll add a test!

seddy:
I've pushed up a test @jbodah. I put it into the existing `guard_spec.rb`, is that OK?

> there didn't appear to be any around syntax highlighting 

Apart from, you know, the `spec/syntax` folder :man_facepalming:! Sorry about that!

spec/syntax/guard_spec.rb
```diff
@@ -2,6 +2,20 @@
...
 require 'spec_helper'
 
+describe 'defguard syntax' do
+  it 'defines `defguard` keyword as elixirGuard' do
+    expect(<<~EOF).to include_elixir_syntax('elixirGuard', 'defguard')
+      defguard some_guard(x) when is_integer(x)
+    EOF
+  end
+
+  it 'defines `defguardb` keyword as elixirPrivateGuard' do
```
jbodah: `defguardb` => `defguardp` 😄

seddy: Gah!

seddy: Fixed.... :crossed_fingers:

PR #409, opened 2018-02-12
Fixes Syntax Highlight

CodyStringham:
This changes the color of `do` and `end` to match `def` - yay.

While I was in there I figured I would fix `if/else/end` as well.


jbodah:
Sounds good to me - thanks!

PR #416, opened 2018-03-30
Install into existing Vim directories

msbit:
Ensure that the files end up in the correct directories if they
already exist.


msbit:
I had an issue where all the files were dropped in my `~/.vim` directory, which looks like it's because the subdirectories already existed. This gets around that, and also works in the case where there are not any existing subdirectories.

PR #422, opened 2018-04-17
Do not set formatprg

lpil:
mix format only knows how to format a full file or module, so using it as formatprg will result in commands using `=` (and similar commands) with any text object that isn't a full module to format incorrectly, or throw a syntax error.

Removing it allows the user to have `=` and similar commands work with the default indenter for some of the file, while still being able to use `mix format` for the full file.

Fixes https://github.com/elixir-editors/vim-elixir/issues/401


jbodah:
I agree with you. Given it's just a one-liner, if someone wants it they can just shove it in their vimrc. I recommend https://github.com/mhinz/vim-mix-format if anyone wants a more complex `mix format` integration

lpil:
Thanks!

PR #424, opened 2018-04-19
Syntax: improve elixirExUnitMacro

### LINKED ISSUE #418: Wrong syntax highlight
Got following type not syntax highlighted as an atom:

```
@type t :: %TestSuite{test: (() -> any), status: :pending}
```

The `test:` key is different from status(from ExUnit I suppose)

mhinz:
Assumption: ExUnit macros are always followed by space.

Technically, a macro's argument can be put into parentheses, but that happens
only seldomly. It's more likely that a function gets named after an ExUnit
macro.

Fixes #418

### DIFF
--- a/syntax/elixir.vim
+++ b/syntax/elixir.vim
@@ -160,7 +160,7 @@ syn match  elixirExceptionDeclaration   "[^[:space:];#<]\+"        contained con
 syn match  elixirCallbackDeclaration    "[^[:space:];#<,()\[\]]\+" contained contains=elixirFunctionDeclaration             skipwhite skipnl
 
 " ExUnit
-syn match  elixirExUnitMacro "\(^\s*\)\@<=\<\(test\|describe\|setup\|setup_all\|on_exit\|doctest\)\>"
+syn match  elixirExUnitMacro "\(^\s*\)\@<=\<\(test\|describe\|setup\|setup_all\|on_exit\|doctest\)\> "
 syn match  elixirExUnitAssert "\(^\s*\)\@<=\<\(assert\|assert_in_delta\|assert_raise\|assert_receive\|assert_received\|catch_error\)\>"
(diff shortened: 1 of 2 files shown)


mhinz:
(The CI passes. At first I just accidentally pushed the branch to this repo directly instead to my fork. That's the failing part.)

syntax/elixir.vim
```diff
@@ -160,7 +160,7 @@ syn match  elixirExceptionDeclaration   "[^[:space:];#<]\+"        contained con
 syn match  elixirCallbackDeclaration    "[^[:space:];#<,()\[\]]\+" contained contains=elixirFunctionDeclaration             skipwhite skipnl
 
 " ExUnit
-syn match  elixirExUnitMacro "\(^\s*\)\@<=\<\(test\|describe\|setup\|setup_all\|on_exit\|doctest\)\>"
+syn match  elixirExUnitMacro "\(^\s*\)\@<=\<\(test\|describe\|setup\|setup_all\|on_exit\|doctest\)\> "
```
jbodah: Another idea would be to use negative lookaround for the `:` character to make sure it's not an atom. I've done that frequently in the indent code (example: https://github.com/elixir-editors/vim-elixir/blob/master/autoload/elixir/indent.vim#L112). That would avoid the parenthesis change

mhinz: But that would leave us with `setup()` beinging highlighted as macro.

Basically I try to distinguish between 3 cases: `test ...`, `test()`, and `test:`. I know the issue is just about the `test:` case, but I guess functions being called `test()` are not too uncommon. 
So.. assuming that the ExUnit macro is always followed by space, we can either add a space at the end of the regex or use that negative lookaround like `\[:\(]\@!`. The outcome is the same.

I guess the question if we think that functions named `test()` are more common than the ExUnit macro `test` with parentheses. (I never saw the latter so far.)

mhinz: Hmm, I somehow missed that response. I agree! Closing this in favor of #432

PR #431, opened 2018-05-10
Reduce `elixirKeyword` syntax group list

hauleth:
Current list seems a little arbitrary about it's content. New one
contains only macros from `Kernel.SpecialForms` plus some macros from
`Kernel` (`if` and `unless`) as broadly known as keywords and some
keyword atoms that are used in `Kernel.SpecialForms`:

 - `else`
 - `after`
 - `rescue`
 - `catch`

As these "feels" like keywords while writing code.

### DIFF
--- a/syntax/elixir.vim
+++ b/syntax/elixir.vim
@@ -18,7 +18,7 @@ syn keyword elixirTodo FIXME NOTE TODO OPTIMIZE XXX HACK contained
 
 syn match elixirId '\<[_a-zA-Z]\w*[!?]\?\>' contains=elixirUnusedVariable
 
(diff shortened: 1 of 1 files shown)


jbodah:
I like this, although I wouldn't mind seeing the guard-friendly functions special-cased again somewhere. Many of those came from [this list](https://hexdocs.pm/elixir/master/guards.html) and I think there's reasonable utility in them (syntax highlight can tell you if it can go in a guard or not so you don't need to check the docs). Still, they aren't keywords (and "keyword" is a poor name for them - I wouldn't mind seeing that change too)

hauleth:
@jbodah the problem is that you can define your own guards and with Elixir 1.6 it became even easier via `defguard(p)` macros, so in that case such colouring would look odd as some calls would be coloured and some would not.

jbodah:
Ah I missed that in the changelogs, good call

PR #435, opened 2018-05-17
Created a dir for UltiSnips snippets and wrote snippets

Afoucaul:
I think it's good to have snippets in such a plugin.
I wrote some for UltiSnips, I suggest that these would be integrated into the plugin.

### DIFF
--- a/UltiSnips/elixir.snippets
+++ b/UltiSnips/elixir.snippets
@@ -1,0 +1,81 @@
+global !p
+import os
+
+def log(*a, **k):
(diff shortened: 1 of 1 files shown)


jbodah:
I feel like somewhere such as https://github.com/honza/vim-snippets would be the conventional place for these and might recommend making a PR there instead. Snippets tend to be very user-specific as opposed to indentation and syntax rules which feel more generic to the language

jbodah:
Closing in favor of pushing these in a common snippet place

PR #444, opened 2018-08-12
Dockerfile and docker-compose setup for local testing

schnittchen:
Make local development more independent of the dev machine setup by providing a docker environment.

- [x] general cleanup
- [x] what about host UID != 1000? (works)
- [x] why does `rspec spec` work but `parallel_rspec spec` fails? (dunno, works now)
- [x] update README

Opening this PR early so more eyes can see this. Would this be a welcome addition to the project?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM ubuntu:latest as base
+RUN useradd -m user
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt install -y bash bundler ruby-rspec xvfb
+RUN DEBIAN_FRONTEND=noninteractive apt install -y vim-gtk3
+RUN mkdir -p /tmp/.X11-unix && chmod 1777 /tmp/.X11-unix
+
+FROM base as user
+USER user
+RUN bundle config path .bundler-path
--- a/README.md
+++ b/README.md
@@ -48,3 +48,12 @@ We've decided not to include `mix format` integration into `vim-elixir`. If you'
 
 Run the tests: `bundle exec parallel_rspec spec`
 Spawn a vim instance with dev configs: `bin/spawn_vim`
+
+Running the tests currently needs `gvim`. If you cannot install a version of vim to get the tests running,
+you can use `docker-compose` to spin up an enviroment for testing. Run
+
+```
+docker-compose run main
+```
+
+This will drop you into a shell where you can run the above commands. Run `docker-compose down` when you are done to free resources.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,32 @@
+version: '3.4'
+services:
+  main:
+    container_name: vim_elixir
+    build:
+      context: .
+      target: user
+
+    volumes:
+      - ./:/home/user/vim-elixir
+      - tmp:/tmp
+
+    stdin_open: true
+    tty: true
+
+    environment:
+      - DISPLAY=:1.0
+    command: bash
+    working_dir: /home/user/vim-elixir
+    depends_on:
+      - xvfb
+
+  xvfb:
+    build:
+      context: .
+      target: user
+    volumes:
+      - tmp:/tmp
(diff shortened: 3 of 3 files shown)


schnittchen:
Can somebody give me early feedback on this? I don't want to invest further work for the case this would not be used and maintained in the future...

jbodah:
@schnittchen sorry for the delay. I'd be okay with merging a docker setup but I'm not sure how much use it would get. My two cents is that knowing vimscript is a bigger blocker than installing gvim and Ruby, but I'm biased as I have a Ruby background myself.

Here are some ways that this could be useful though:

* Getting the tests off of gvim and onto vim-server. I haven't been able to get the tests to pass without gvim/macvim and it would be nice to not have to rely on the GUI

* Adding support for parallel testing. The vim server instance is currently global, and the tests are pretty slow as they involve a lot of IO and chatter with the server instance. It would be really nice to have support for parallel testing

* Extracting a generic test harness. In general I think a lot of the work done in this project around testing could be useful to other projects going forward. For example, we have a test framework setup (rspec), a bunch of useful test helpers (check syntax, check indent, check indent while typing), and some generic tooling (benchmarking, loading a fresh vim instance with your project's files, etc). It would be nice to pull all of this out into something that could be used and maintained by a larger audience. Docker could fit in by giving the plugin maintainer an isolated environment

* Fixing the flaky build system. CI relies on gvim and has a tendency to crap out at times. Encapsulating the tests in a docker instance could add some reliability

So TL;DR - I don't think a purely development-focused docker setup is going to get used (and I may be wrong) as most devs already have vim installed and most contributors seem experienced/okay with letting CI give them feedback. There are a few other places where I could see a docker build being useful though, and I would support merging this if it could help with some of the above ideas. Otherwise I feel like it's not going to be used (please comment if you disagree)

schnittchen:
I think your first, second and third point are somewhat orthogonal. All of this can potentially be easier to hack on with a Docker setup.

For me having a docker setup would tremendously help here because I do not want to mess with my development system. Also I find that the Dockerfile could serve to some degree as a documentation of what currently works, setup-wise. Also, I don't like waiting for CI but would rather execute a single test or test file on a keypress.

I have some experience running Docker in CI (but not with travis). The problem I see here with the current state of my PR is the build times, because CI usually won't let you use a docker cache. Maybe we can find a better base image -- OTOH this could become easier when gvim is no longer needed.

schnittchen:
Questions:

* why do we rely on gvim?
* do we potentially have features that rely on gvim, so we need to stick with that?
* "Extracting a generic test harness" sound useful if there is serious overlap. This could mean rspec utility code, a Docker setup, even both. What do you have in mind?

jbodah:
> why do we rely on gvim?

This predates me a bit. I've tried to get things working with vim server without much luck. I don't think we have any specific features that rely on gvim, but I've had issues crop up when trying to switch it (can be changed here: https://github.com/elixir-editors/vim-elixir/blob/master/spec/spec_helper.rb#L200)

> "Extracting a generic test harness" sound useful if there is serious overlap. This could mean rspec utility code, a Docker setup, even both. What do you have in mind?

I think a container that had Ruby + vimrunner + rspec + our rspec helpers would be a great start. Then any language project could just fork that project and have a nice test harness ready to go. They'd be able to start throwing in examples for indentation and syntax by just adding specs that use our rspec helpers

schnittchen:
I got this a little further to enable me to work on folding testing. The workflow is like this:

```
docker-compose run main
```

This will build the image if it hasn't already, then launch two connected containers (one for Xvfb) and enters a shell in the "main" one. There one can run

```
bundle exec rspec spec # or
bundle exec parallel_rspec spec
```

The Dockerfile is much tidier now.

schnittchen:
I've mentioned the steps necessary when using docker in the Readme. Should I add a caveat that that workflow is not necessary well maintained?

jbodah:
Closing; please use the Docker setup currently in the repo

PR #451, opened 2018-11-02
Add basic projectionist support

mluts:
Hello, few days ago i bothered to add support for https://github.com/tpope/vim-projectionist to my config files. Basically i use `A` shortcut to navigate between tests/source-code, i think it could be useful for other people.

Feel free to modify, thank you for this repo.

### DIFF
--- a/ftplugin/elixir.vim
+++ b/ftplugin/elixir.vim
@@ -51,3 +51,19 @@ silent! setlocal formatoptions-=t formatoptions+=croqlj
 
 let b:undo_ftplugin = 'setlocal sw< sts< et< isk< com< cms< path< inex< sua< def<'.
       \ '| unlet! b:match_ignorecase b:match_words b:block_begin b:block_end'
+
+function s:projectionist_detect_elixir()
+  if filereadable('mix.exs')
+    let root = simplify(fnamemodify('.', ':p:s?[\/]$??'))
+
+    let projections = {}
+    let projections['lib/*.ex'] = { 'type': 'main', 'alternate': ['test/{}_test.exs'] }
+    let projections['lib/**/*.ex'] = { 'type': 'main', 'alternate': ['test/{}_test.exs'] }
+    let projections['test/*_test.exs'] = { 'type': 'test', 'alternate': ['lib/{}.ex'] }
+    let projections['test/**/*_test.exs'] = { 'type': 'test', 'alternate': ['lib/{}.ex'] }
+
+    call projectionist#append(root, projections)
+  endif
+endfunc
+
+autocmd User ProjectionistDetect call s:projectionist_detect_elixir()


jparise:
You could also set `make` to `mix` (for source files) or `exunit` (for test files).

ftplugin/elixir.vim
```diff
@@ -51,3 +51,19 @@ silent! setlocal formatoptions-=t formatoptions+=croqlj
 
 let b:undo_ftplugin = 'setlocal sw< sts< et< isk< com< cms< path< inex< sua< def<'.
       \ '| unlet! b:match_ignorecase b:match_words b:block_begin b:block_end'
+
+function s:projectionist_detect_elixir()
+  if filereadable('mix.exs')
+    let root = simplify(fnamemodify('.', ':p:s?[\/]$??'))
+
+    let projections = {}
+    let projections['lib/*.ex'] = { 'type': 'main', 'alternate': ['test/{}_test.exs'] }
```
jparise: I prefer `type: source` to `type: main`, which follows the convention used by the Projectionist documentation. What do you think?

ftplugin/elixir.vim
```diff
@@ -51,3 +51,19 @@ silent! setlocal formatoptions-=t formatoptions+=croqlj
...
+    let projections['lib/**/*.ex'] = { 'type': 'main', 'alternate': ['test/{}_test.exs'] }
```
jparise: Are these additional `lib/**/` and `test/**/` entries necessary? The `lib/*.ex` / `test/*.ex` entries work well for me, even with file hierarchies.

hauleth: No, these aren't needed, quoting docs:

> From a globbing perspective, `"*"` is actually a stand in for `"**/*"`.  For advanced cases, you can include both globs explicitly: `"test/**/test_*.rb"`. When expanding with `{}`, the `**` and `*` portions are joined with a slash.  If necessary, the `dirname` and `basename` expansions can be used to split the value back apart.

ftplugin/elixir.vim
hauleth: I use `type: lib`

hauleth:
My configuration for projectionist looks like this:

```viml
let g:projectionist_heuristics['mix.exs'] = {
            \ 'apps/*/mix.exs': { 'type': 'app' },
            \ 'lib/*.ex': {
            \   'type': 'lib',
            \   'alternate': 'test/{}_test.exs',
            \   'template': ['defmodule {camelcase|capitalize|dot} do', 'end'],
            \ },
            \ 'test/*_test.exs': {
            \   'type': 'test',
            \   'alternate': 'lib/{}.ex',
            \   'template': ['defmodule {camelcase|capitalize|dot}Test do', '  use ExUnit.Case', '', '  alias {camelcase|capitalize|dot}, as: Subject', '', '  doctest Subject', 'end'],
            \ },
            \ 'mix.exs': { 'type': 'mix' },
            \ 'config/*.exs': { 'type': 'config' },
            \ '*.ex': {
            \   'makery': {
            \     'lint': { 'compiler': 'credo' },
            \     'test': { 'compiler': 'exunit' },
            \     'build': { 'compiler': 'mix' }
            \   }
            \ },
            \ '*.exs': {
            \   'makery': {
            \     'lint': { 'compiler': 'credo' },
            \     'test': { 'compiler': 'exunit' },
            \     'build': { 'compiler': 'mix' }
            \   }
            \ }
            \ }
```

kelvinst:
@hauleth I like your configuration, 👍 on configuring it that way... I would just add the dispatch config to it, since it's a well known package too

hauleth:
@kelvinst no problem, however I use a little bit more powerful [makery](https://github.com/igemnace/vim-makery). That is why there is no Dispatch configuration.

kelvinst:
Yep, realized that, will check that one someday, but for now I'm sticking with dispatch.

kelvinst:
I opened a PR for @mluts branch there: https://github.com/mluts/vim-elixir/pull/1. The changes should fall here once he accepts (I hope, never did that before).

sodapopcan:
This is quite old and doesn't look like there are any signs of it getting merged but just wanted to weigh in here:

I don't think it's a good idea for this to be included in this repo for the very reason outlined by the above convo and comments on the PR: `projections.json` files are very personal.  I feel that anyone who has installed projectionist already has their own system for including their projections.

I think this type of thing would be best included under a "Tips" category or the like in the wiki or README.

PR #453, opened 2018-11-25
Folding tests

schnittchen:
Here are a very basic number of folding tests, including a failing (skipped) test for https://github.com/elixir-editors/vim-elixir/issues/375


jbodah:
These look excellent, thank you for looking into it. I will look into this deeper over the next day or so (currently away from home) and will merge

jbodah:
Looks great - thanks again!

PR #457, opened 2018-12-12
Fix syntax highlighting with "~S" sigil

feliperenan:
Fixes https://github.com/elixir-editors/vim-elixir/issues/438

Just changed to use the same pattern used in `elixirString`
https://github.com/elixir-editors/vim-elixir/blob/34873ee66e0dfd7c8105d6f0ad048a2df04e77f2/syntax/elixir.vim#L77

### Before
image

image

### After
image

image

### DIFF
--- a/spec/syntax/doc_spec.rb
+++ b/spec/syntax/doc_spec.rb
@@ -35,6 +35,17 @@ describe 'documentation syntax' do
       expect(ex).to include_elixir_syntax('elixirDocString', 'foo')
     end
 
+    it 'doc with sigil_S triple double-quoted multiline content with parentheses' do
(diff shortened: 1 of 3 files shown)


jbodah:
Thanks, looks great! Would you mind adding a test for this?

Here's an example: https://github.com/elixir-editors/vim-elixir/blob/master/spec/syntax/doc_spec.rb#L27-L36

I'll run those prior on my machine and merge if it all looks good

feliperenan:
@jbodah Tests added. I'm not sure if they are covering the scenarios correctly though, could you please take a look?

jbodah:
Looks good - thanks again!

PR #467, opened 2018-12-30
Use 'set filetype' to set filetype once again

jparise:
The `s:setf()` helper function was originally introduced to avoid setting the filetype multiple times (375e9e1). I believe this covered the case where a file had an `.ex` or `.exs` extension and also include "elixir" in the shebang line (such that `s:DetectElixir()` would match and call `set filetype` redundantly).

c56bd82 removed the existing `&filetype` check in `s:setf()`, relying on the `&filetype` global and ftdetect framework to avoid redundant updates.

I believe we can simplify this even further by using just `set filetype` in the file extension-based autocommands and including a guard against (re)setting filetype in `s:DetectElixir()`.

Also, use explicit case-insensitive comparison operators. This isn't a good place to make assumptions about `ignorecase`.


jbodah:
Sorry for the delay on this. Must've missed it during the holidays

PR #470, opened 2019-02-12
Syntax: Add doc delimiters

endersstocker:
Why
---

I want to be able to differentiate between ordinary delimiters and those used for docs

How
---

* Add **elixirDocSigilDelimiter** group
* Add **elixirDocStringDelimiter** group
* Link **elixirDocStringDelimiter** and **elixirDocStringDelimiter** to **elixirStringDelimiter** and **elixirSigilDelimiter** to preserve existing behavior

### DIFF
--- a/spec/syntax/doc_spec.rb
+++ b/spec/syntax/doc_spec.rb
@@ -5,11 +5,15 @@ require 'spec_helper'
 describe 'documentation syntax' do
   describe 'string' do
     it 'doc in double quotes' do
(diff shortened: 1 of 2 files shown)


jbodah:
Looks fine to me - would you mind adding a simple test to make sure we don't regress on this if we refactor the syntax rules? There should be some examples in https://github.com/elixir-editors/vim-elixir/blob/master/spec/syntax/doc_spec.rb

endersstocker:
@jbodah Updated

jbodah:
Thanks!

PR #481, opened 2019-03-08
vim-dadbod integration for applications using Ecto

hauleth:
[vim-dadbod][1] is useful library for interacting with databases, this
PR adds support for interacting with applications that are using Ecto
with `Ecto.Adapter.Postgres` and `Ecto.Adapter.MySQL`, thanks to that
user can run:

    :DB ecto:MyApp.Repo SELECT * FROM foo

And it will open new preview window containing result of the query or
even one can use

    :DB ecto:MyApp.Repo

To open interactive shell for DB assigned to given repository.

[1]: https://github.com/tpope/vim-dadbod

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,8 @@ Features:
 * Syntax highlighting for Elixir and EEx files
 * Filetype detection for `.ex`, `.exs` and `.eex` files
 * Automatic indentation
+* Integration between Ecto projects and [vim-dadbod][] for running SQL queries
+  on defined Ecto repositories
 
 ## Installation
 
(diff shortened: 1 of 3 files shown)


jbodah:
Looks good; could you add instructions to the README?

skbolton:
@hauleth I think your comment on how to use it needs to go in the readme or docs. I had to find this issue in order to learn how to do the dadbod integration

hauleth:
@skbolton I will update docs after weekend.

oliverswitzer:
@skbolton Hi, was this ever added to the docs? I can't seem to find anything about it.

hauleth:
@oliverswitzer I do not think so. The only documentation is the mention in the README. I will try to write something today.

PR #484, opened 2019-03-21
Add LiveEEx template (.leex) syntax support

crunchysoul:
Add Phoenix LiveEEx Template (`.leex`) file type support.
Add LiveView sigil (`~L"""LiveView"""`) support.

Below is a comparison screenshot: 
compare_with_current

PR #485, opened 2019-03-21
Syntax highlighted properly for *.leex files

RomanKotov:
Small changes to the logic of choosing embedded elixir subtype.


jbodah:
This has some overlap with #484. I'd like to get that one merged as it has a few other fixes with it as well. Leaving this on deck for now

RomanKotov:
Ok, sure.

jbodah:
This should be fixed by #490

PR #487, opened 2019-03-22
Add Phoenix LiveEEx Template (.leex) file type support.

crunchysoul:
@jbodah  I closed the original #484 , please check this new pull quest, thank you.

Add Phoenix LiveEEx Template (`.leex`) file syntax highlight support.
Add LiveView sigil (~L"""LiveView""") syntax highlight support in `.ex` file.
Update README.md

Screen shot of `.leex` syntax highlight:

Screen Shot 2019-03-22 at 4 43 05 pm

Screen shot of LiveView sigil syntax highlight in `.ex`:

Screen Shot 2019-03-22 at 4 44 57 pm

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -122,6 +122,12 @@ module EexBuffer
   end
 end
 
+module EexBuffer
+  def self.new
+    Buffer.new(VIM, :leex)
+  end
+end
+
 RSpec::Matchers.define :be_typed_with_right_indent do |syntax|
   buffer = Buffer.new(VIM, syntax || :ex)
 
@@ -145,6 +151,7 @@ end
 {
   be_elixir_indentation:  :ex,
   be_eelixir_indentation: :eex
+  be_eelixir_indentation: :leex
 }.each do |matcher, type|
   RSpec::Matchers.define matcher do
     buffer = Buffer.new(VIM, type)
@@ -170,6 +177,7 @@ end
 {
   include_elixir_syntax:  :ex,
   include_eelixir_syntax: :eex
+  include_eelixir_syntax: :leex
 }.each do |matcher, type|
   RSpec::Matchers.define matcher do |syntax, pattern|
     buffer = Buffer.new(VIM, type)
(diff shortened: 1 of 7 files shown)


spec/spec_helper.rb
```diff
@@ -145,6 +151,7 @@ def self.new
 {
   be_elixir_indentation:  :ex,
   be_eelixir_indentation: :eex
+  be_eelixir_indentation: :leex
```
jbodah: I think this should read:

```
{
{
  be_elixir_indentation:  :ex,
  be_eelixir_indentation: :eex,
  be_leelixir_indentation: :leex
}
```

spec/spec_helper.rb
```diff
@@ -170,6 +177,7 @@ def self.new
 {
   include_elixir_syntax:  :ex,
   include_eelixir_syntax: :eex
+  include_eelixir_syntax: :leex
```
jbodah: Similar for syntax

spec/syntax/sigil_spec.rb
```diff
@@ -87,6 +87,11 @@
     it 'without escaped parenthesis' do
       expect('~S(\( )').not_to include_elixir_syntax('elixirRegexEscapePunctuation', '( ')
     end
+
+    it 'Live EEx' do
+      expect('~L"""liveview template"""').to include_elixir_syntax('elixirSigilDelimiter', '""""')
+      expect('~L"""liveview template"""').to include_elixir_syntax('elixirSigilDelimiter', '""""')
```
jbodah: Are these expects different? Could we just have one?

jbodah: If possible it would be nice to get a test on the inner sigil syntax being recognized. I can see if I can get one written

spec/spec_helper.rb
crunchysoul: thank you for pointing out, fixed.

spec/spec_helper.rb
crunchysoul: thank you for pointing out, fixed.

spec/syntax/sigil_spec.rb
crunchysoul: I removed the duplicated line, but haven't got time to write test for inner sigil syntax yet.

PR #491, opened 2019-04-25
ftdetect: Fix detection when reopening

liskin:
When doing :e in an already opened elixir script lacking the .exs filename suffix, we would skip the check and end up with filetype=conf or something like that.
    
Fix this by forcing the check when `!did_filetype()` even if `&filetype ==# 'elixir'`.

### DIFF
--- a/ftdetect/elixir.vim
+++ b/ftdetect/elixir.vim
@@ -3,7 +3,7 @@ au BufRead,BufNewFile *.eex,*.leex set filetype=eelixir
 au BufRead,BufNewFile * call s:DetectElixir()
 
 function! s:DetectElixir()
-  if &filetype !=# 'elixir' && getline(1) =~# '^#!.*\<elixir\>'
+  if (!did_filetype() || &filetype !=# 'elixir') && getline(1) =~# '^#!.*\<elixir\>'
     set filetype=elixir
   endif
 endfunction


ftdetect/elixir.vim
```diff
@@ -1,9 +1,9 @@
-au BufRead,BufNewFile *.ex,*.exs set filetype=elixir
-au BufRead,BufNewFile *.eex,*.leex set filetype=eelixir
+au BufRead,BufNewFile *.ex,*.exs setfiletype elixir
+au BufRead,BufNewFile *.eex,*.leex setfiletype eelixir
```
mhinz: This will lead to new files opened as `euphoria3` instead of `elixir`: https://github.com/elixir-editors/vim-elixir/issues/475

liskin: Oh, I'm sorry, I should've git blamed the code a bit. I'll drop that bit. Thanks a lot for catching this! :+1:

mhinz:
I can't seem to reproduce the original issue.

I can use `:e` in an already opened `elixir` buffer just fine. Maybe the issue is a bit more complicated or related to another plugin or something?

liskin:
@mhinz The problem is specifically with elixir scripts not named with the `.exs` suffix. I'll clarify that in the commit message.

liskin:
So I did the git blame now and saw all the iterations it went through, so I had to implement it in another way to preserve `set ft` but prevent setting it twice.

mhinz:
Can you please give the steps to reproduce the actual issue this is supposed to solve?

I thought this was about starting out with a file `test`, then adding the shebang and doing `:e` to get the filetype set. But that works on master as well. What am I missing?

As far as I see, `ft=conf` is only set when the first line starts with `#!` without Vim detecting a better filetype.

liskin:
@mhinz The first `:e` after adding the shebang is fine, but try it for a second time.
(My usecase was opening an elixir script, changing `b:ale_linters`, then `:e` to make that change do something.)

liskin:
Anyway, exact repro steps:

1. `$ echo '#! /usr/bin/env elixir' >test-script`
2. `$ vim test-script`
3. `:set ft` (outputs `filetype=elixir`)
4. `:e`
5. `:set ft` (outputs `filetype=conf`)

mhinz:
If the shebang is `#!elixir`, then, with the same steps, it still gets set to `conf`.

IMHO, it's Vim being wrong here. `:au filetypedetect BufRead` says:

```
    *         if !did_filetype() && expand("<amatch>") !~ g:ft_ignore_pat    && (getline(1) =~ '^#' || getline(2) =~ '^#' || getline(3) =~ '^#'^I|| getline(4) =~ '^#' || getline(5) =~ '^#') |   setf FALLBACK conf | endif
```

IMHO, this should include another `empty(&filetype)` condition...

PR #492, opened 2019-04-30
syntax: Two alias-related fixes

liskin:
* highlight dots in `Foo.Bar.fun` the same as in `defmodule Foo.Bar`
* highlight uppercased atoms in keyword lists (`[ Atom: 1 ]`) as atoms


jbodah:
Thanks!

PR #493, opened 2019-05-14
syntax: Prevent syntax/{html,javascript}.vim from overriding our syn sync

liskin:
When including other syntax types it's necessary to declare that we're
the main syntax, otherwise they will apply their own syn sync settings.

Here, by including html.vim in 7e00da6033b7847c6bb71df18f852342946eab42,
we end up with syn sync minlines=10 maxlines=100 and we sync using
html.vim syncing rules, which isn't at all what we want. Try opening
large_file.ex in the middle, syntax will be incorrect.

Unfortunately the main_syntax trick doesn't prevent javascript.vim from
doing syn sync fromstart. This is most likely a bug in javascript.vim
and I'll submit a pull request to vim to fix this. Luckily setting
main_syntax does prevent javascript.vim from introducing its own syncing
rules, it just overrides minlines and maxlines, so we can work around it
by moving the syn sync minlines to the end.

### DIFF
--- a/syntax/elixir.vim
+++ b/syntax/elixir.vim
@@ -8,9 +8,6 @@ endif
 let s:cpo_save = &cpo
 set cpo&vim
 
-" syncing starts 2000 lines before top line so docstrings don't screw things up
-syn sync minlines=2000
-
 syn cluster elixirNotTop contains=@elixirRegexSpecial,@elixirStringContained,@elixirDeclaration,elixirTodo,elixirArguments,elixirBlockDefinition,elixirUnusedVariable,elixirStructDelimiter
 syn cluster elixirRegexSpecial contains=elixirRegexEscape,elixirRegexCharClass,elixirRegexQuantifier,elixirRegexEscapePunctuation
 syn cluster elixirStringContained contains=elixirInterpolation,elixirRegexEscape,elixirRegexCharClass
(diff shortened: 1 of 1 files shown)


liskin:
vim PR, for reference: https://github.com/vim/vim/pull/4372

jbodah:
@liskin thanks for this. Would you be able to come up with repro steps so I can better understand what's happening? I tried opening `large_file.ex` in the middle but I didn't see the issue

liskin:
@jbodah `vim +161 large_file.ex` triggers it here.

Also:

```
:syn sync
--- Syntax sync items ---
syncing on items; minimal 10, maximal 100 lines before top line
htmlHighlight  xxx match +<[/a-zA-Z]+  contained groupthere NONE
…
```

jbodah:
Thanks - was able to repro with vim 8 but not nvim

liskin:
That's a bit weird, nvim has the same syntax files for html and javascript. Maybe there's some hack in it that makes `main_syntax` unnecessary? Dunno. :-/

jbodah:
Closing for now; if there is enough interest/pain in this then we can merge it but I'd prefer to see this fixed at a lower level

liskin:
@jbodah Well, feel free to drop the javascript hack commit then. The `main_syntax` thing is not a hack, it's how syntax scripts are meant to interact with each other.

liskin:
(Also worth noting that the maintainer of `syntax/javascript.vim` is not responding to my emails so the likelihood of that bit being fixed “at a lower level” this year is lower than I originally expected.)

jbodah:
Going forward with this for now. We can redo this in the future if we care

PR #495, opened 2019-06-27
Set mix.lock as Elixir

dbernheisel:
This explicitly sets `mix.lock` as an Elixir file. Otherwise, the file loads with no detected file format.


jbodah:
Great - thanks!

PR #506, opened 2019-08-29
Add syntax highlight to function calls

w-sanches:
# Description
Function call highlight was [added to elixir syntax highlight on Sublime Text](https://github.com/elixir-editors/elixir-tmbundle/pull/179) and I thought that was a neat addition as it makes it easier to scan what are function vs variables in the code. I propose we can also add this to our Vim syntax highlight.

# Result
image

### DIFF
--- a/spec/syntax/function_spec.rb
+++ b/spec/syntax/function_spec.rb
@@ -48,13 +48,25 @@ describe 'function syntax' do
     EOF
   end
 
+  it 'detects function calls appended by atom with parenthesis' do
+    expect(<<~'EOF').to include_elixir_syntax('elixirFunctionCall', 'func')
+      :mod.func()
+    EOF
+  end
+
   it 'detects function calls appended by module without parenthesis' do
     expect(<<~'EOF').to include_elixir_syntax('elixirFunctionCall', 'func')
       Mod.func
     EOF
   end
 
-  it 'does not highlight terms without a parentheresis' do
+  it 'detects function calls appended by atom without parenthesis' do
+    expect(<<~'EOF').to include_elixir_syntax('elixirFunctionCall', 'func')
+      :mod.func
+    EOF
+  end
+
+  it 'does not highlight function calls without parameters that have no parenthesis' do
     expect(<<~'EOF').not_to include_elixir_syntax('elixirFunctionCall', 'func')
       func
     EOF
--- a/spec/syntax/function_spec.rb
+++ b/spec/syntax/function_spec.rb
@@ -18,15 +18,57 @@ describe 'function syntax' do
     EOF
   end
 
+  it 'detects higher order function calls' do
+    expect(<<~'EOF').to include_elixir_syntax('elixirFunctionCall', 'func')
+      func.()
+    EOF
+  end
+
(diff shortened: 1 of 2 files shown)


spec/syntax/function_spec.rb
```diff
@@ -17,4 +17,16 @@
         __ensure_defimpl__(protocol, for, env)
     EOF
   end
+
+  it 'detects function calls with parenthesis' do
+    expect(<<~'EOF').to include_elixir_syntax('elixirFunctionCall', 'func')
+      func()
```
CaiqueMitsuoka: WDYT adding coverage to more cases.
The examples on the PR description brings lots of them, but the main one that I think it's missing is when the call has a parenthesis and has a `.` like `func.()`.

jbodah:
I like this. The only things I'm not sure of:

* The default highlight group is set to `Function`. This seems odd to me - I'd expect colorschemes to bind to `elixirFunctionCall` to get this behavior and for the default to be unhighlighted

* Let's add more tests, particularly some of the interesting ones like ones with bangs, spaces, no parens, etc

w-sanches:
@CaiqueMitsuoka @jbodah just added more tests, please LMK if I forgot some important test.

I did set the default group to `Function` just so they are consistently coloured throughout the file (either call or definition, they are still functions).
One can always set different colours to `elixirFunctionCall` and `elixirFunctionDefinition` if they wish and overwrite this behaviour. IIRC this is the behaviour for C files, but not the same for Ruby so I don't think there's some consistency through different languages.

spec/syntax/function_spec.rb
```diff
@@ -17,4 +17,58 @@
...
+
+  it 'detects function calls appended by module without parenthesis' do
+    expect(<<~'EOF').to include_elixir_syntax('elixirFunctionCall', 'func')
+      Mod.func
+    EOF
+  end
+
+  it 'does not highlight terms without a parentheresis' do
+    expect(<<~'EOF').not_to include_elixir_syntax('elixirFunctionCall', 'func')
+      func
```
jbodah: This seems odd and I don't know what the solution here is. This causes only some function calls to be highlighted by default:

image

Both `hd` instances come back as `elixirId` rather than `elixirFunctionCall`

w-sanches: I have to confess I completely forgot we could use atoms as module names. Just pushed a commit to fix that.

Also, while at it I realised `|> func` should also be highlighted since if we are piping it is not ambiguous that we are calling a function.

jbodah:
I want to give this a bit more manually testing; will get back to this by end of week

denvaar:
Can this be merged?

jbodah:
The only thing I'm iffy on in this branch is that functions without args and parens aren't detected which feels misleading. I am okay to merge this as long as we match these to `elixirFunctionCall` and don't link this to `Function` by default (due to the above inconsistency). Users can define a link themselves if they want this functionality and are okay with the inconsistency. Macros are also a weird grey area (especially structural ones like `if` and `with`)

The original branch was removed, so I will copy this over to a new branch, remove the link, and merge

PR #511, opened 2019-09-03
Revert "Fix highlight for atoms containing @ character"

jbodah:
Reverts elixir-editors/vim-elixir#508


jbodah:
This commit had some failing specs; reverting

PR #512, opened 2019-09-03
Fix highlight for atoms containing @ character

### LINKED ISSUE #510: eb06df6 breaks highlighting
After applying eb06df6, syntax highlighting appears to be mostly broken for me.

Example from `mix.exs`:

Bildschirmfoto 2019-09-03 um 20 17 07

I'm using the latest Neovim commit (`NVIM v0.4.0-1944-g38806f23e`).

Please let me know if you need more information.

Tagging @w-sanches, since he authored the commit.

w-sanches:
Fixes #510 

In eb06df6 I used a regexp that was matching too many things. I completely forgot that spaces are also printable characters 🤦‍♂ 

This PR comes with a new version using `[^\d0-\d127]` to match those special characters. This will:

- **Not** match spaces, so atoms will be matched correctly;
- Match accented characters like `á ó é`;
- Match least common characters like `ಠ` which **are** valid in an atom like `:ಠ‿ಠ`;
- Match **some** invalid characters;

Elixir actually accepts any UTF-8 letter as atoms and Vim does not support Unicode modifiers AFAIK, so using something like `\p{L}` is not possible 😞 
I went with the route of adding any non-ASCII on top of words and @s as it gave us more positive results and the false-positives will raise errors when compiled, so they are easily fixable. I still can't decide by myself if we should add those false-positives or not.
WDYT?

The other option would be to remove support for accents and other UTF-8 characters and only highlight those atoms with `\w` and `@` matched chars.

### DIFF
--- a/spec/syntax/atom_spec.rb
+++ b/spec/syntax/atom_spec.rb
@@ -88,4 +88,31 @@ describe 'Atom syntax' do
     end
     EOF
   end
+
+  it 'detects atoms containing @ in it' do
+    expect(<<~EOF).to include_elixir_syntax('elixirAtom', '@somewhere')
+      :atom@somewhere
+    EOF
+    expect(<<~EOF).to include_elixir_syntax('elixirAtom', '@somewhere')
+      [atom@somewhere: nil]
+    EOF
+  end
+
+  it 'detects atoms containing Unicode letters in it' do
+    expect(<<~EOF).to include_elixir_syntax('elixirAtom', 'ó')
+      :atóm
+    EOF
+  end
+
+  it 'does not incorrectly detects spaces as part of the atom' do
+    input = <<~EOF
+      {:power_assert, "~> 0.2.0", only: :test}
+    EOF
+    expect(input).to include_elixir_syntax('elixirAtom', 'power_assert')
+    expect(input).to include_elixir_syntax('elixirAtom', 'only')
+    expect(input).to include_elixir_syntax('elixirAtom', 'test')
+    expect(input).not_to include_elixir_syntax('elixirAtom', '0.2.0')
+    expect(input).not_to include_elixir_syntax('elixirAtom', '{')
+    expect(input).not_to include_elixir_syntax('elixirAtom', '}')
+  end
 end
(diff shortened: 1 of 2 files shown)


sodapopcan:
Is the only thing stopping this from getting merged the conflicts, or is there something wrong with the commit?  I'd be happy to fix the conflicts.

samanera:
> Vim does not support Unicode modifiers AFAIK

I think `[[:lower:][:upper:]]` should work with Unicode.

> `:ಠ‿ಠ`

I got this error:

```
** (SyntaxError) iex:1:2: unexpected token: "‿" (column 2, code point U+203F)
    |
  1 | :ಠ‿ಠ
    |  ^
    (iex 1.15.6) lib/iex/evaluator.ex:294: IEx.Evaluator.parse_eval_inspect/4
    (iex 1.15.6) lib/iex/evaluator.ex:187: IEx.Evaluator.loop/1
    (iex 1.15.6) lib/iex/evaluator.ex:32: IEx.Evaluator.init/5
    (stdlib 5.0.2) proc_lib.erl:241: :proc_lib.init_p_do_apply/3
```

Elixir's [syntax reference](https://github.com/elixir-lang/elixir/blob/main/lib/elixir/pages/references/syntax-reference.md#atoms) for atoms mentions that unquoted atoms should start with _unicode letters or an underscore_ and then it supports numbers and at symbols (`@`).

I think the following regex can match unquoted atoms: `:[_[:lower:][:upper:]][_[:lower:][:upper:][:digit:]@]*`

I would use a `syn-region` for quoted atoms, although, I would need to think a bit more about this approach.

samanera:
I would defined it as:

```vim
syn region qatoms start=/:"/ skip=/\\"/ end=/"/
syn region qatoms start=/:'/ skip=/\\'/ end=/'/
syn match  uatoms /:[_[:lower:][:upper:]][_[:lower:][:upper:][:digit:]@]*[?!]\?/

syn cluster atoms contains=qatoms,uatoms

hi link qatoms identifier
hi link uatoms identifier
```

This is just an idea, I might send a patch later 🙂 .

PR #517, opened 2020-02-10
Add ctags to gitignore

pksunkara:
I submodule this repo to use it with pathogen and it keeps breaking when I update because of untracked content in the submodule. This should fix the issue


jbodah:
Seems fine

PR #521, opened 2020-03-11
add elixirFunctionCall

jbodah:
Implementation from https://github.com/elixir-editors/vim-elixir/pull/506

cc: Thanks @w-sanches


jbodah:
Note function/macro calls will no longer be `elixirId`; `elixirId` will only be used for variable names now

PR #537, opened 2020-11-26
Add Surface support and EEX sigil support

### LINKED ISSUE #534: EEx sigil "~E" html does not have highlighting
The html within EEx `~E` sigils does not appear to be highlighting, but LiveView sigils look to work just fine.

# Actual

Screen Shot 2020-10-20 at 5 56 33 PM

# Expected

Screen Shot 2020-10-20 at 5 55 56 PM
Screen Shot 2020-10-20 at 5 56 51 PM

I added this line to `syntax/elixir.vim` and it seems to fix the issue, but not sure if it's the appropriate way to do it. I copied the line used for LiveView sigils, and just replaced the `L`.
```
syntax region elixirTemplateSigil matchgroup=elixirSigilDelimiter keepend start=+\~E\z("""\)+ end=+^\s*\z1+ skip=+\\"+ contains=@HTML fold
```

dbernheisel:
This adds support for *.sface files, introduced by [Surface](https://hex.pm/packages/surface) ([demo](http://surface-demo.msaraiva.io/getting_started)). This is much like the existing embedded elixir types, but instead of `<% ... %>` for interpolation, it uses `{{ ... }}`, and uses the `~H` sigil when in Elixir files..

When adding the `~H` support, I noticed that `~E` and `~e` were not included, and found this reported in #534. I decided to include that here as well, so this PR also resolves #534 

Before:
image

After:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 Features:
 
 * Syntax highlighting for Elixir and EEx files
-* Filetype detection for `.ex`, `.exs`, `.eex` and `.leex` files
+* Filetype detection for `.ex`, `.exs`, `.eex`, `.leex`, and `.sface` files
 * Automatic indentation
 * Integration between Ecto projects and [vim-dadbod][] for running SQL queries
   on defined Ecto repositories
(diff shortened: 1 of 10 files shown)


jbodah:
LGTM thanks!

Issue #536, opened 2020-11-13
LiveEEX sigil "~L" does not switch to EEx-style indentation


whitepaperclip:
Running `ggVG=` to indent the file leads to:

# Actual

```ex
defmodule MyApp.Foo do
  use MyApp.Web, :live_component

  @impl true
  def mount(socket) do
    {:ok, socket}
  end

  @impl true
  def render(assigns) do
    ~L"""
    <div>
    <p>component!</p>
    </div>
    """
  end
end

```

# Expected

```ex
defmodule MyApp.Foo do
  use MyApp.Web, :live_component

  @impl true
  def mount(socket) do
    {:ok, socket}
  end

  @impl true
  def render(assigns) do
    ~L"""
    <div>
      <p>component!</p>
    </div>
    """
  end
end

```

alappe:
I do have the same issue… :-/ Makes editing the LiveView files rather tedious.

PR #538, opened 2021-02-28
Auto-indent embedded views

sodapopcan:
I'm submitting this PR as a request for feedback.  It aims to solve the indentation issues with embedded views (stuff between `~L`, `~H`, `~e`, and `~E`) as per #536.  I couldn't figure out how to embed indent syntax—and assumed this is probably not possible—so this is what I've come up with.  The aim is to provide an improved experience over the current behaviour, not to write a full re-implementation HTML indentation.

First and foremost, while I've added some test cases, I cannot for the life of me figure out how to run them.  `$ bundle exec parallel_rspec spec` has every test failing with `Vimrunner::NoSuitableVimError`.  `bin/vim` boots up a vim instance but I'm not sure how to target it.

Otherwise, there are a couple of outstanding issues I'm aware of:
- There is some funky behaviour when using [vim-surround](tpope/vim-surround) on a multi-line opening tag, though this seems to be current behaviour of the plugin.
- [close-tag](/alvan/vim-closetag)'s `>` insert mapping leaves the cursor at the same indentation level as the opening tag.
- `/>` on a line by itself is caught by the `top_of_file` handler.  Moving my proposed handler to the opening position fixes this, but I'm currently worried about doing so since I can't run the tests.

Thank you!

### DIFF
--- a/autoload/elixir/indent.vim
+++ b/autoload/elixir/indent.vim
@@ -20,6 +20,7 @@ function! elixir#indent#indent(lnum)
   call cursor(lnum, 0)
 
   let handlers = [
+        \'inside_embedded_view',
         \'top_of_file',
         \'starts_with_string_continuation',
         \'following_trailing_binary_operator',
@@ -63,6 +64,17 @@ endfunction
 
 function! s:prev_starts_with(context, expr)
   return s:_starts_with(a:context.prev_nb_text, a:expr, a:context.prev_nb_lnum)
+endfunction
+
+function! s:in_embedded_view()
+  let groups = map(synstack(line('.'), col('.')), "synIDattr(v:val, 'name')")
+  for group in ['elixirPhoenixESigil', 'elixirLiveViewSigil', 'elixirSurfaceSigil']
+    if index(groups, group) >= 0
+      return 1
+    endif
+  endfor
+
+  return 0
 endfunction
 
 " Returns 0 or 1 based on whether or not the text starts with the given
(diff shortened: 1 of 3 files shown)


jbodah:
> I couldn't figure out how to embed indent syntax—and assumed this is probably not possible

vim-ruby's ERB has some examples of this. https://github.com/vim-ruby/vim-ruby/blob/4788a08433c3c90e131fc7d110d82577e1234a86/indent/eruby.vim#L36 is where the new indentexpr is defined for ERB, and it is composing the HTML and Ruby indentation expressions. It seems a little awkard however that we might possibly have a cycle here (Elixir embeds EEx which embeds Elixir)

As long as we have tests for anything we implement, it doesn't really matter that much to me what the implementation of this looks like (assuming we can't further write breaking tests), so doing it this way is better than being unsupported altogether, and tests should keep us from breaking users in the future

> First and foremost, while I've added some test cases, I cannot for the life of me figure out how to run them. $ bundle exec parallel_rspec spec has every test failing with Vimrunner::NoSuitableVimError

I just updated the README with some additional information. The test suite depends on a GUI vim (gvim, mvim). The easiest solution is to have one of these in your PATH, however I also added a simple dotfile for specifying the path if the binary lives elsewhere. Please let me know if that works for you or not

> Otherwise, there are a couple of outstanding issues I'm aware of

I'm not super familar with each of these - would you be able to take a snapshot of this behavior? If it isn't too serious then I think it's okay to merge even if these integrations are broken (I generally take the stance that we do not support integrations first-class so that we can get larger features like this merged). We can create follow-up issues for tracking if these are minor

jbodah:
Just ran this locally - the tests are passing for me with the exception of the new tests. The `i` helper generates two test cases - one which tests that "select all-indent" works properly and another that tests indent is correct when typed. I think this will be fine if we remove the latter case

Here's an example test that only implements the first case:

https://github.com/elixir-editors/vim-elixir/blob/527e6fd8798638a79621e0b5c788b67b2b4b4dbc/spec/indent/documentation_spec.rb#L14-L24

autoload/elixir/indent.vim
```diff
@@ -65,6 +66,19 @@ function! s:prev_starts_with(context, expr)
...
 
+function! s:in_embedded_view()
+  let position = getpos('.')
+  let lnum = line('.')
+  let start_lnum = search('\s\+\~\%\(e\|E\|L\|H\)"""', 'bW')
+  if start_lnum == 0
+    return 0
+  endif
+  let end_lnum = search('\s\+"""', 'W')
+  call setpos('.', position)
```
jbodah, line 77: This moves the cursor - is this necessary? We do reset the cursor between handler calls so it's not a huge deal, but I'd be curious what this enables

spec/indent/embedded_views_spec.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'Indenting anonymous functions' do
+  i <<~EOF
+  def render(assigns)
```
jbodah, line 7: These are missing the trailing `do`; could you add that?

autoload/elixir/indent.vim
```diff
@@ -355,6 +369,39 @@ function! s:do_handle_pattern_match_block(relative_line, context)
   end
 endfunction
 
+function! elixir#indent#handle_inside_embedded_view(context)
+  if s:in_embedded_view()
```
jbodah, line 373: We might be able to simplify this with a syntax check, e.g.

Given

```elixir
def render(assigns) do
  ~H"""
  <Component
    id="some-id"
    value={{ @some_value}}
  >
    Inner HTML
  </Component>
  """
end
```

With `synstack(3, 3)` (note: cursor pos) we get: `[145, 396, 163, 157]`. Those can be mapped to the respective syntax group. `synIDattr(396, "name")` returns `elixirSurfaceSigil`. So we can dump the stack and search if we're nested in an embedded view this way

autoload/elixir/indent.vim
jbodah: (Note this might be superseded by the syntax comment)

jbodah:
Thanks for contributing this! So far it looks good. I'd like to give the actual indent rules a deeper read, but it looks correct so far

sodapopcan:
Hi, thanks so much for the review and feedback!

I have now been able to get the tests running but it seems no matter what indentation I put, they pass.
```ruby
  it "indents" do
    expect(<<~EOF).to be_eelixir_indentation do # doesn't matter if I use this or `be_elixir_indentation`
    ~e"""
    <div>
      <p>
        Some text
        </p>
          </div>
    """
    EOF
    end
  end
```

The above passes.  Do you know why that would be?

Thanks for the tip about `synstack()`!  I've refactored to use it.  Previously, there was certainly no need to move the cursor, the `n` flag just slipped my mind.

In general, this was a pretty rushed PR.  I decided I was going to solve this problem on Saturday and I got overly-excited when everything started working so I made a PR the same day.  I would like to keep it as a draft for another week or so while I test it out and work out some kinks, and I'm hoping some others will try it as well.  And yes, that'd be great if you could think on the rules.  I've already made some changes that I haven't pushed yet but will do that at some point today.

Thanks again!

sodapopcan:
> I'm not super familar with each of these - would you be able to take a snapshot of this behavior? If it isn't too serious then I think it's okay to merge even if these integrations are broken (I generally take the stance that we do not support integrations first-class so that we can get larger features like this merged). We can create follow-up issues for tracking if these are minor

Close tag auto-closes html tags.  If you type a second `>`, it will put them on separate lines with the cursor in betweem.  The problem I'm seeing is fairly minor:
```html
  <div>
  | <!-- cursor ends up here, inline with the tags instead of indented
  <div>
```

I can't even begin to describe what happens with vim-surround but suffice it to say it's an existing problem with multi-line opening tags in any markup file.

sodapopcan:
Ok, I've pushed (I also intend to squash all these before requesting to merge).

I moved the embedded rules to the top.  This solved a few issues I was having:
- Pressing `<cr>` in insert on after an opening sigil would make the sigil fly off to the right
- A solitary `/>` would end up heavily indented to the right (it was being caught by `starts_with_binary_operator`)
- ...this was yesterday and I can't remember the third thing 😬 

I think in the end a few of the branches can be collapsed into one since they all have the same body, though it certainly is helpful having them separated for debug purposes.

sodapopcan:
Also, off-topic question:  I'm working on syntax highlighting for Surface—would that belong in this repo or its own?

sodapopcan:
So, I restarted this because a _whole_ whack of stuff wasn't working in my original PR.  I've spent so much time on this over the past week that I still haven't had a lot of time to actually use it.  I've added several test cases though the code itself is pretty hard to reason about.  I realized there was no way I was going to get away with simply parsing HTML with `searchpair()` so it's turned a complex.  It could probably be cleaned up a bit and I'll be looking at that.  The last thing I have to do is make sure all Surface directive stuff works.  Otherwise, I just wanted to give an update and solicit some more feedback.

Even if this remains complex, I would love it if this could be merged and hopefully get more eyes on this.  I will definitely be available to support as my care level that this works is very high :)

jbodah:
> Also, off-topic question: I'm working on syntax highlighting for Surface—would that belong in this repo or its own?

Here is fine. I've become more lax about including things outside of base Elixir now that this plugin is fairly stable

sodapopcan:
I've pulled this out of draft mode and squashed into one commit.  The logic is pretty complex and can probably be simplified but I just wanted some feedback and I'm trying to get people to try it out.  I've been using it the past week and fixing things along the way.  Other than the complexity, the rules seem pretty solid.  I have some pretty decent tests so I can definitely try and simplify it, I'm just a little burnt out on it atm.

vinniefranco:
@sodapopcan awesome work on this! hopefully it's merged in soon

sodapopcan:
@vinniefranco Thank you, although I have not had time to work on this in quite some time.  For a while there, I was spending all my hobby time on programming which led to some burn out.  I've since been focusing on non-programming hobbies and since I don't write elixir professionally, I haven't had the motivation to get back to this.  It's been simmering in the back of my mind but I have no idea when I'll be getting back to it.  I feel my approach is a little brute force.  It works quite well with Surface templates (and I think these style of templates are now in LiveView... am I mistaken?) but still has a few (perhaps severals) cases where it messes up standard `leex` templates.  All this to say that I'm sorry, I don't know when this will be complete!  It could be quite a while unless someone can swoop in and take over.

whitepaperclip:
@sodapopcan I rebased your branch off of the upstream master and can confirm it's still working really well!

jbodah:
Tests are passing; Going to merge this - we can follow-up on issues via bug reports

sodapopcan:
Great, thanks for merging!  I'm actually just getting back into a LiveView project using HEEx so I'll be testing this out a bunch.

PR #544, opened 2021-04-29
Added char lists syntax highlighting

### LINKED ISSUE #545: Charlists syntax highlighting
I believe that char lists and strings could have a different syntax group, since they are very different data types.
Even if both had the same color as your regular color scheme's String group, just giving different names would make it easier for someone willing to customize its own colors.
Pull request #544 is a possible implementation that I came up with for this issue.

lucassperez:
I believe that strings and charlists deserve different highlighting group, since those are very different data types.
Even if it is defaulted to regular vim string, it would make a little easier for us to customize our own colors.

**The motivation behind this** is that I ran into a bug when I was pattern matching the wrong value because I didn't notice I wrote single quoted "strings" instead of double quoted strings by mistake, and different colors for them would have saved me some minutes and a couple of head scratches.

I made this change today and I'm not sure they are 100% safe and did not break anything. I do not excel in those syntax groups of vim, but it seems to be working.
Comments and changes to my piece of code are more than welcome, and if this change is not what we want/need, then we can close this pull request.

Thanks for reading and providing us with this good content! (:

fixes #545

### DIFF
--- a/autoload/elixir/indent.vim
+++ b/autoload/elixir/indent.vim
@@ -104,7 +104,7 @@ endfunction
 " Returns 0 or 1 based on whether or not the given line number and column
 " number pair is a string or comment
 function! s:is_string_or_comment(line, col)
-  return s:syntax_name(a:line, a:col) =~ '\%(String\|Comment\)'
+  return s:syntax_name(a:line, a:col) =~ '\%(String\|Comment\|CharList\)'
 endfunction
 
 function! s:syntax_name(line, col)
--- a/spec/indent/basic_spec.rb
+++ b/spec/indent/basic_spec.rb
@@ -558,4 +558,12 @@ describe 'Basic indenting' do
     :send
   end
   EOF
+
+  i <<~EOF
+  defmodule Hi do
+    def hello_world do
+      "end"
+      'end'
+    end
+  EOF
 end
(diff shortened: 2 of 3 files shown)


jbodah:
I think this is a good idea. The only thing I'm not sure of here is whether or not this will affect indent rules which ignore special characters (e.g. `do`/`end`) which are strings: https://github.com/elixir-editors/vim-elixir/blob/c3cb96e153728fbfd050173b4af19118b131f278/autoload/elixir/indent.vim#L148

lucassperez:
I'm sorry, I'm not sure what is the rule we are talking about here, but I tried to type these snippets in my vim:
```ex
defmodule Hi do
  |
```
```ex
defmodule Hi do
  def hello do
    "Hello"
    |
```
```ex
defmodule Hi do
  def hello do
    "Hello"
  end
  |
```
The `|` symbols are the position of the cursor after I pressed enter in each of those situations. It seems normal so far?

One maybe questionable decision is using `Identifier` here, since it is the same group of things such as `self`, `@doc` and module attributes in general:
https://github.com/elixir-editors/vim-elixir/blob/a7cf8c9b1f36cea03b6889ef6e81530f8e188f53/syntax/elixir.vim#L233
https://github.com/elixir-editors/vim-elixir/blob/a7cf8c9b1f36cea03b6889ef6e81530f8e188f53/syntax/elixir.vim#L238

If we change it to `String` would this solve your concern? Strings and char lists would have the same color by default, but at least it would be easy for each person to customize their own.

jbodah:
I pushed up a fix for the indent code

I think we should link `elixirCharListDelimiter` to `Delimiter` and `elixirCharList` to `String` (so they are highlighted like strings by default); WDYT?

lucassperez:
I think it is a very good ideia. I agree that linking them to String by default seems like the most reasonable idea, and it still makes it very easy to customize into any other color if someone wants to.

The bug fix is also great :clap:

PR #546, opened 2021-05-14
Encode all possible DB options

hauleth:
Previous implementation was buggy in several ways:

- It required `:username`, `:password`, `:hostname`, and `:database` fields while only `:database` is required by Ecto
- It ignored `c:init/2` in the repo definition
- It ignored additional options passed to the repository
- Did not allowed for `:socket_dir` configuration

This fixes all these problems at once.


jbodah:
LGTM

PR #547, opened 2021-05-15
Update script to install plugin manually

anthonylgf:
- I updated the script to install de plugin to follow [the official method to install plugins in VIM](https://github.com/vim/vim/blob/03c3bd9fd094c1aede2e8fe3ad8fd25b9f033053/runtime/doc/repeat.txt#L515).
    - I added an option to the user to choose the directory name where the plugin files will be installed inside the `~/.vim` directory.

### DIFF
--- a/manual_install.sh
+++ b/manual_install.sh
@@ -1,9 +1,64 @@
 #!/usr/bin/env bash
 
-mkdir -p ~/.vim/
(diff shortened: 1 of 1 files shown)


jbodah:
Seems good to me, thanks

PR #549, opened 2021-06-12
Remove no affect replace

marocchino:
this codes from https://github.com/tpope/vim-abolish/blob/master/plugin/abolish.vim#L120-L127

```viml
    let word = substitute(word,'\(\u\+\)\(\u\l\)','\1_\2','g')
```

line was duplicated.

```viml
    let word = substitute(word,'-','_','g')
```

elixir modules dose not allows kebab-case module name so we do not need it.

### DIFF
--- a/autoload/elixir/util.vim
+++ b/autoload/elixir/util.vim
@@ -15,9 +15,7 @@ function! elixir#util#get_filename(word) abort
 
   " convert to snake_case
   let word = substitute(word,'\(\u\+\)\(\u\l\)','\1_\2','g')
-  let word = substitute(word,'\(\u\+\)\(\u\l\)','\1_\2','g')
   let word = substitute(word,'\(\l\|\d\)\(\u\)','\1_\2','g')
-  let word = substitute(word,'-','_','g')
   let word = tolower(word)
 
   return word


jbodah:
Is this a hard rule? I see [here](https://hexdocs.pm/elixir/1.12/naming-conventions.html) that underscore-style is a soft convention

marocchino:
It will be ArithmeticError.
```ex
> Kebab-Case
** (ArithmeticError) bad argument in arithmetic expression: Kebab - Case
    :erlang.-(Kebab, Case)
```

It may be possible to declare by putting an atom directly, but it can be ignored because aliases are not available. 

```ex
> defmodule :"Aaa.Bbb.Kebab-Case" do
> end
 <<70, 79, 82, 49, 0, 0, 3, 244, 66, 69, 65, 77, 65, 116, 85, 56, 0, 0, 0, 137,
   0, 0, 0, 13, 18, 65, 97, 97, 46, 66, 98, 98, 46, 75, 101, 98, 97, 98, 45, 67,
   97, 115, 101, 8, 95, 95, 105, ...>>, nil}

> alias :"Aaa.Bbb.Kebab-Case"
:"Aaa.Bbb.Kebab-Case" # expect :"Kebab-Case"
> alias Aaa.Bbb.:"Kebab-Case"
** (SyntaxError) iex:12:15: syntax error before: 'Kebab-Case'
```

PR #551, opened 2021-07-29
Add *.lexs to ftdetect for Temple templates

rktjmp:
Adds support for [Temple](https://github.com/mhanberg/temple) templates.

They are just elixir code, so all that is needed is to match the extra extension: `*.lexs` (`*.exs` is already covered obviously).


mhanberg:
Just saw this!

In the mean time, I have this in my init.lua config `autocmd([[BufRead,BufNewFile *.lexs set filetype=elixir]])`

jbodah:
LGTM thanks

PR #560, opened 2022-03-27
Remove ftdetect

mhanberg:
Elixir and all of the most popular template syntaxes
(LEEx, HEEx, Surface) are included in Vim and Neovim's native filetype
detection and now longer have to provided by a plugin.

I noticed the FileType event was being emitted twice when opening Elixir
files, which was causing me some problems. This should fix that.

This also switches the .lexs extension for eelixir to elixir, as that
those files are have elixir in them, not eelixir.

### DIFF
--- a/ftdetect/elixir.vim
+++ b/ftdetect/elixir.vim
@@ -1,6 +1,4 @@
-au BufRead,BufNewFile *.ex,*.exs set filetype=elixir
-au BufRead,BufNewFile *.eex,*.heex,*.leex,*.sface,*.lexs set filetype=eelixir
-au BufRead,BufNewFile mix.lock set filetype=elixir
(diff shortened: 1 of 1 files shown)


adamu:
The docs say

> Filetype detection for .ex, .exs, .eex, .heex, .leex, and .sface files

Do they need to be updated?

sodapopcan:
I'm actually still confused by the file type detection here.  I've had it manually added to my vimrc based on [this discussion](https://github.com/elixir-editors/vim-elixir/issues/562) for a long time now.  I remember I tried to remove it maybe a year ago and it broke stuff, but I just tried removing now and everything seems to be highlighting correctly so ya, that line should probably be removed?

PR #561, opened 2022-03-28
Add defn syntax highlighting

### LINKED ISSUE #550: Numerical definitions (Nx/Defn) don't have syntax highlighting
Minor issue, but I just noticed using [numerical definitions from nx](https://github.com/elixir-nx/nx/tree/main/nx#numerical-definitions) breaks syntax highlighting for function definitions.

# Actual

No syntax highlighting for `defn`-defined function:

Screen Shot 2021-07-09 at 8 14 33 AM

# Expected

Syntax highlighting matches that of normal function:

Screen Shot 2021-07-09 at 8 14 49 AM

These screenshots are from a kitty terminal using neovim 0.5.0 stable and the [gruvbox colorscheme](https://github.com/gruvbox-community/gruvbox).

duff:
Fixes #550.


jbodah:
I would generally recommend non-native libraries come up with their own support or users use their own `after/` files, but this seems harmless enough. Thanks for the fix

grzuy:
For what is worth, also adding numerical private defs in https://github.com/elixir-editors/vim-elixir/pull/568.
Thanks!

PR #570, opened 2024-06-16
Improve embedded HEEx and Surface syntax highlighting

### LINKED ISSUE #566: ~H sigil syntax highlighting breaks on < and > inside embedded elixir tags
Working theory is that the syntax highlighting for Elixir (.ex) files is not looking for the `%` sign to determine a matching closing tag for embedded Elixir opening tags. This means a `>` character would prematurely close the embedded tag while a `<` character would appear to open a tag within a tag. 

# Actual
image
image

# Expected
```ex
<%= if Enum.count(@assets) > 1 do %>
```
```ex
<%= for asset <- @assets do %>
```

sodapopcan:
Currently, embedded HEEx is simply highlighted as HTML.  This breaks as soon as you have lines like `:for={i <- @numbers}` or `:if={a > 2}` where the `<` and `>` are treated as HTML.  There are also other random things like `{[` that completely break highlighting.  I've been just silently dealing with this for four years now and figured I'd fix it anyway I could!

I saw the problem was mainly that you end up with a recursion error when you have `syntax/elixir.vim` calling `syn import @EELIXIR syntax/eelixir.vim` and `syntax/eelixir.vim` calling `syn import @elixirTop syntax/elixir.vim`.  After trying trying to hack around this, I figured duplication was the best answer and it's worked quite nicely.

This PR does this following:

* Properly highlight embedded HEEx (and Surface)
* Re-name groups to favour HEEx naming (this would break any customizations people have done)
* Add highlight group for `phx-` attributes
* Add highlight group for heex attribute (eg, `some_attr="some value"`)
* Add highlight group for heex special attribute (ie, `:if`, `:for`, and `:let`)

I have not copied this stuff over to `syntax/eelixir.vim` as I wanted a review first.  I can certainly roll back any of the breaking change stuff, but I don't know how many Elixirists still use vanilla Vim.

Thanks!

Closes #566

### DIFF
--- a/syntax/elixir.vim
+++ b/syntax/elixir.vim
@@ -109,15 +109,41 @@ syn region elixirSigil matchgroup=elixirSigilDelimiter start="\~\l\/"           
 syn region elixirSigil matchgroup=elixirSigilDelimiter start=+\~\a\z("""\)+ end=+^\s*\z1+ skip=+\\"+ fold
 syn region elixirSigil matchgroup=elixirSigilDelimiter start=+\~\a\z('''\)+ end=+^\s*\z1+ skip=+\\'+ fold
 
-
-" LiveView Sigils surrounded with ~L"""
+" LiveView-specific sigils for embedded templates
 syntax include @HTML syntax/html.vim
 unlet b:current_syntax
-syntax region elixirLiveViewSigil matchgroup=elixirSigilDelimiter keepend start=+\~L\z("""\)+ end=+^\s*\z1+ skip=+\\"+ contains=@HTML fold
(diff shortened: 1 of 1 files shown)


sodapopcan:
This is ready.

I can't for the life of me figure out how to contain region delimiters within other regions.  Is this even possible?  In any event, template delimiters get matched inside elixir code, ie, outside of template sigils.  This was mainly a problem for heex's `{}` as it was matching tuples.  I got around this requiring a `=` before the `start` match.

jbodah (approved):
I have not had a chance to test this, however we can just iterate if there's follow-up

sodapopcan:
Thank you!

PR #573, opened 2024-07-06
Add `g:loaded_elixir`

sodapopcan:
This is a very simple commit that adds and checks `g:loaded_elixir`.

I'm developing a plugin that partially depends on this one and I have found no other reliable way to detect this plugin has been loaded.  I've added it to `plugin/` so that it is set at load time as opposed to autoload time to avoid load order bugs.

Thanks!

### DIFF
--- a/plugin/elixir.vim
+++ b/plugin/elixir.vim
@@ -1,0 +1,4 @@
+if exists("g:loaded_elixir")
+  finish
+endif
+let g:loaded_elixir = 1


jparise:
This change appears to break [vim-plug's on-demand loading mechanism](https://github.com/junegunn/vim-plug?tab=readme-ov-file#on-demand-loading-of-plugins).

This `Plug` call:

```vim
Plug 'elixir-editors/vim-elixir', {'for': 'elixir'}
```

... was previously sufficient to lazily load this plugin for both direct Elixir buffers (filetype `elixir`) _as well as_ indirect usages, such as embedding the `elixir` syntax in fenced markdown code blocks via `g:markdown_fenced_languages = ['elixir']`.

The latter no longer works unless you also load vim-elixir for markdown buffers (`Plug 'elixir-editors/vim-elixir', {'for': ['elixir', 'markdown']}`).

I _think_ the issue is more about the `plugin/elixir.vim` file itself rather than the `g:loaded_elixir` guard within it. I still see the issue if the file is entirely empty.

I don't have an alternate solution to propose, but it seems like this current approach is exposing some kind of flaw.

sodapopcan:
Interesting!  I need to read up more on this but I'm totally happy to have this reverted.

Thanks for speaking up @jparise and apologies for messing with your workflow.

jbodah:
Reverted in #575

PR #581, opened 2024-09-12
Update HEEx comment syntax

sodapopcan:
This commit prefers the present-day comment syntax for HEEx files.

Looking for a lil' code review here if anyone is up for it otherwise I will merge in a few days.

It's not perfect mainly in that when commenting out a tag like `<a>` which has its own highlighting, the comment highlighting will be off unless the whole tag is commented out.  Similarly, commenting inside `<script>` and `<style>` get no special handling.  No errors are shown, though, which makes it a nice step up from status-quo.

### DIFF
--- a/ftplugin/eelixir.vim
+++ b/ftplugin/eelixir.vim
@@ -108,8 +108,8 @@ if !exists('b:surround_5')
   let b:surround_5 = "<% \r %>\n<% end %>"
 endif
 
-setlocal comments=:<%#
-setlocal commentstring=<%#\ %s\ %>
+setlocal comments=:<%!--
+setlocal commentstring=<%!--\ %s\ --%>
 
 let b:undo_ftplugin = "setl cms< " .
       \ " | unlet! b:browsefilter b:match_words | " . s:undo_ftplugin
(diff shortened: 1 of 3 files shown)


jbodah:
I haven’t used embedded Elixir enough to be of much help here, however there are testing primitives in this repo for ensuring syntax is correct if you wanted to blast this with tests

jbodah:
Which platform are you developing on? I can help you get them running. You'll need (1) gvim or macvim and (2) Ruby to start

sodapopcan:
Thanks for the offer and sorry for the delayed response here.  I'm on Mac.  I got it working fine a couple of years ago when I made my first PR and wrote a bunch of tests.  I'll try it out again this week.

It's great that you bring up getting it off Ruby as I had the same thought.  This would effectively kill three dependencies (ruby, docker, and the requirement for a gvim).  Unfortunately vim test libs are moving targets.  Currently it looks like [themis](https://github.com/thinca/vim-themis) is the only one that is maintained.

sodapopcan:
This actually ended up getting rolled into another commit a while ago.

PR #582, opened 2024-09-18
[WIP] port tests to run completely in vim

jbodah:
This PR is working towards porting all of the tests to run directly in vim (I have only tested with nvim). It may be difficult to get some of the typing tests implemented, however paste-then-indent tests work well (and are much faster)

```
$ ./run_vimtest
```

Failures open up a buffer with something like:

image

Making the indent tests directory based means we can simply import Elixir code and get free testing.


run_vimtest
```diff
@@ -0,0 +1,3 @@
+#! /usr/bin/env sh
+set +xe
+vim -c ":so test.vim"
```
jbodah, line 3: Need to clear out the vimrc and initialize it to something predictable

vimtest/syntax/alias3.ex
```diff
@@ -0,0 +1,3 @@
+# empty
```
jbodah, line 1: First line = search term; second line = syngroup; rest = body tested on

sodapopcan:
This is amazing!  I can take a look on the weekend.  I still regular Vim so I can test there and help with conversions.

One thing after a quick read, other test frameworks often `:new` for each test, `setlocal` for options, the `:close`.  This may be a simpler way to handle clean up.

jbodah:
Have you come across any other test suites? I’d be curious to see examples of what other projects have done. I agree creating new buffers would be simpler. I’d also be interested in getting a more formal test runner (specifically something with exit codes that would could get running in CI). I tried themis a bit but struggled getting anything meaningful tested and couldn’t figure out how to log to stdout

sodapopcan:
Oh boy, I forgot about this one!  Sorry for never getting back to you.

In short, I have been using [Vader.vim](junegunn/vader.vim) in other projects although I'm unsure about exit codes.  I may or may not take this up myself if I feel the motivation.  It's largely been very low since it seems the vast majority of the Elixir community, at the vocal Elixir community, use NeoVim without this plugin.

